PR #221, opened 2017-08-03
Clean up Gemfile; only use NewRelic in production

awead:
Puts New Relic in production only, so we don't run into problems locally, then configures Travis to run our CI build and check for Rubocop violations; lastly, fix the outstanding violations! 🚓

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
+
 # lock '3.6'
 
 # Set assets roles to occur on jobs as well as web
-set :assets_role, [:web, :job]
+set :assets_role, %i[web job]
 
 # application and repo settings
 set :application, 'cho'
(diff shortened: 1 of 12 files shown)


awead:
@cam156 I think this PRs is stepping a bit on your PR's toes, but I need to do this to finish setting up Travis. It was failing when I submitted other PRs. I think if you rebase yours against this, it should work. The Gemfile will be very different because I re-arranged everything for clarity. Hope that's ok.

cam156:
@awead I have travis about half set up on my box, so we were both working on the same thing.  WAS there a reason not to merge my PR? I do not see any comments

config/deploy.rb
```diff
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
+
 # lock '3.6'
 
 # Set assets roles to occur on jobs as well as web
-set :assets_role, [:web, :job]
+set :assets_role, %i[web job]
```
cam156: We need to figure out how to get rubocop running the correct style guide version

cam156: if this PR gets merged reverting the rubocop comit and rerunning rubocop should work: https://github.com/psu-stewardship/guides/pull/3/files

cam156:
@awead I am going to dismiss my review so it does not block you while I am away.

PR #262, opened 2017-09-08
Commented out `newrelic_rpm` as it is causing conflicts and our subsc…

informaticianme:
…ription is inactive


cam156:
@informaticianme This PR should have included the Gemfile.lock also.  Here is some documentation on Bundler: https://bundler.io/v1.3/rationale.html

Issue #15, opened 2016-02-08
System wide metadata defaults


bmrgoldman:
As a metadata specialist, I would like cross-repository metadata templates so that I can apply metadata requirements across the whole repository.

cam156:
Required fields across collections.  Each collection can have additional items, but these requirements must be present

cam156:
This is related to the Base Metadata Application Profile (MAP)  that @lmballinger is working on.  How do we enforce the information in MAP.

cam156:
https://psu.app.box.com/file/211247382460 - Is the content DM MAP.  Any required fields here should be required in CHO.  The CHO MAP will be larger.

cam156:
Related to #57 & #16.  The fields define here will be the one people can choose from for the templates.

System Metadata Workflow: https://psu.app.box.com/notes/218648875937

cam156:
This will include defining what autocompletes are available to the user in the template creation.

cam156:
* Assume we have a FieldValidatorFactory that takes the validation type and returns a class that can perform that validation.

* Assume we have a FieldPresentorFactory that takes the display transformation name and returns a class that can transform the value to return the presented value.

* Assume we have a ControlledVocabularyFactory that takes the controlled vocabulary name and returns a controlled vocabulary

* Field Type: string, text, date, number (are there other types)

* Required: required, recommended, optional

```
rails generate scaffold MetadataField label:string field_type:integer required:integer
validation:string multiple:boolean controlled_vocabulary:string default_value:string display_name:string display_transformation:string
```

cam156:
- [x] Create a model object to store the information
- [ ] CRUD Operations
- [ ] Export as CSV
- [ ] Import CSV
- [ ] FieldValidatorFactory
- [ ] FieldPresentorFactory
- [ ] ControlledVocabularyFactory

awead:
This is now called the data dictionary, see #323

PR #269, opened 2017-09-27
Adding a model to store a singular field in the Base MAP. refs #15

cam156:
This is the basic model for storing information about a field in the Base MAP (Metadata Application Profile)

Each field can have the following specified:
  * label
  * field type
  * requirement designation
  * controlled vocabulary
  * default value
  * display name
  * display transformation
  * multiple
  * validation

### DIFF
--- a/app/models/concerns/with_field_type.rb
+++ b/app/models/concerns/with_field_type.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+module WithFieldType
+  extend ActiveSupport::Concern
+
+  included do
+    enum field_type: { string: 'string',
+                       text: 'text',
+                       numeric: 'numeric',
+                       date: 'date' }
+  end
+end
--- a/app/models/concerns/with_requirement_designation.rb
+++ b/app/models/concerns/with_requirement_designation.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+module WithRequirementDesignation
+  extend ActiveSupport::Concern
+
+  included do
+    enum requirement_designation: { required_to_publish: 'required to publish',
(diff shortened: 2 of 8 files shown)


spec/models/metadata_field_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+                              'default_value' => 'abc123',
+                              'display_name' => 'My Abc123',
+                              'display_transformation' => 'abc123_transform',
+                              'field_type' => 'date',
+                              'id' => model.id,
+                              'label' => 'abc123_label',
+                              'multiple' => false,
+                              'requirement_designation' => 'recommended',
+                              'updated_at' => model.updated_at,
+                              'validation' => 'abc123_validation' } }
```
awead: @cam156 would this ever be a hash with symbols as keys? I think parameter hashes use strings, so that should be fine. I often get confused and find `HashWithIndifferentAccess` helpful, but that doesn't need to be a requirement here.

awead:
@cam156 looks great! Left one question, and just needs a rebase.

spec/models/metadata_field_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    expect(model.requirement_designation).to eq('optional')
+  end
+
+  it 'can be text' do
+    model.text!
+    expect(model.field_type).to eq('text')
+  end
+
+  context 'with metadata' do
+    let(:expected_metadata) { { 'controlled_vocabulary' => 'abc123_vocab',
```
cam156: @awead I moved the hash into the context since it is only used here.  I do not think it will ever have symbols, since it is the output of the .attributes method

PR #271, opened 2017-09-29
Adding in classes from running generate scaffold

cam156:
Allows for basic CRUD of MetadataApplicationProfileFields within the application


app/controllers/metadata_application_profile_fields_controller.rb
```diff
@@ -0,0 +1,92 @@
...
+  def create
+    no_error = false
+    begin
+      metadata_application_profile_field = MetadataApplicationProfileField.new(metadata_field_params)
+      no_error = metadata_application_profile_field.save
+    rescue ArgumentError => e
+      no_error = false
+      metadata_application_profile_field = MetadataApplicationProfileField.new
+      metadata_application_profile_field.errors.add('Invalid argument', e.message)
+    end
```
awead: I thought the convention was that `save` would just return false, with errors added to the AR object. Is there something about `MetadataApplicationProfileField` that raises errors instead of rendering validation messages?

cam156: @awead yes.  The validation on the Field_type enumerations and Required_designation enumerations produces argument errors when the object is created.

cam156:
@awead this PR is back up and passing.  Now using the Factory.  Looks like it ups our coverage significantly. (Maybe just becuase there is more code)

PR #278, opened 2017-10-02
adding in presenter for csv view

cam156:
This PR adds a presenter that allows the index to be viewed (exported) as csv.

Note there were no changes to the controller to allow for the additional format.  Rails forwards to the correct views based on format.

If we like this pattern of thin controller  and view that uses a presenter we may want to refactor the index views for the json code that was auto generated for us to use a presenter also. That said the code that was generated is fairly slick in that is uses the same partial for a single item and for each item in the array.


app/views/metadata_application_profile_fields/index.csv.erb
```diff
@@ -1,3 +1,2 @@
-# frozen_string_literal: true
-
-json.array! @metadata_application_profile_fields, partial: 'metadata_application_profile_fields/metadata_field', as: :metadata_application_profile_field
+<%- presenter = ::CsvMetadataFieldListPresenter.new(@metadata_application_profile_fields) %>
```
awead: Can the controller create the presenter instead?

cam156: @awead so the controller is not handeling the different output types in the scenario, but rails is pointing at the correct view.  Following the example that was generated for json I just added new views and Viola a new format was supported with no changes to the controller.  So the answer is yes we could put logic into the controller to create the correct presenter based on format, or we could move it to a helper, or just leave it here.  I know we do not want much code in the views, but given that it is a simple one liner I was ok putting it here.

This is different from the hyrax model, but I was wondering if maybe that was OK.

awead: @cam156 I think the ideal solution would be a MAPFieldIndexPresenter, that serve both the regular index view and the csv one. Something like:
``` ruby
def index
  @metadata_application_profile_fields = MAPFieldIndexPresenter.new(MetadataApplicationProfileField.all)
end
```
And this view becomes just: `<%= @metadata_application_profile_fields.to_csv %>`
However, we can save this for a later refactoring, if you'd like.

cam156: @awead why would we want one presenter for serving up all formats?  That seems to be violating single responsibility.  I present everything seems less understandable than I present a CSV view of the world.  What happens  to the everything presenter when we start converting to different XML formats?  I was putting the code closest to where it was used until we had some repeatable pattern that we could use to logically refactor.   I'm certainly willing to change the controller to put in switch if that is what you would prefer, but I am not on board with creating an everything presenter.

awead: That's a fair point, but I think a presenter's responsibility is knowing how to format the data, even in multiple ways. However, you could delegate all the different formats to sub-presenters, or services. That does make for easier testing. My point was that all the logic for dealing with formatting the data is in one place, and the presenter knows which formats are available. We don't have to make this change now, though. Wait and see when we have xml in its multiple varieties and see if this still makes sense.

PR #279, opened 2017-10-03
Adding csv import service

cam156:


spec/services/metadata_application_profile_csv_importer_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  let(:importer) { described_class.new(file) }
+
+  it 'parses a file' do
+    expect {
+      importer.import
+    }.to change(MetadataApplicationProfileField, :count).by(1)
+  end
+
+  context 'record exists' do
+    let(:field) { create :metadata_application_profile_field }
```
awead: Use `let!` here and you should be able to avoid the before call.

cam156: @awead rubocop does not enjoy let, so I started doing it this way.  Is rubocop now allowing let!

cam156: spec/services/metadata_application_profile_csv_importer_spec.rb:20:5: C: RSpec/LetSetup: Do not use let! for test setup.
    let!(:field) { create :metadata_application_profile_field }
    ^^^^^^^^^^^^

awead: weird!

PR #281, opened 2017-10-04
Enable deleting works from the UI

awead:
Provides the standard Rails delete link to remove a work object from the
system. This will delete it from both Solr and Postgres.

### DIFF
--- a/app/controllers/work_objects_controller.rb
+++ b/app/controllers/work_objects_controller.rb
@@ -53,7 +53,7 @@ class WorkObjectsController < ApplicationController
     change_set_persister.buffer_into_index do |buffered_changeset_persister|
       buffered_changeset_persister.delete(resource: change_set)
     end
-    flash[:alert] = "Deleted #{change_set.resource}"
+    flash[:alert] = "#{change_set.title.first} has been deleted"
     redirect_to(root_path)
   end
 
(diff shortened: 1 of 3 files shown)


app/controllers/work_objects_controller.rb
```diff
@@ -53,7 +53,7 @@ def destroy
     change_set_persister.buffer_into_index do |buffered_changeset_persister|
       buffered_changeset_persister.delete(resource: change_set)
     end
-    flash[:alert] = "Deleted #{change_set.resource}"
+    flash[:alert] = "#{change_set.title.first} as been deleted"
```
cam156: @awead guessing you wanted `has` instead of `as` here.

spec/features/edit_work_object_spec.rb
```diff
@@ -23,4 +23,12 @@
       expect(page).to have_content("Title can't be blank")
     end
   end
+
+  context 'when deleting a work' do
+    it 'removes the work from the system' do
+      visit(edit_work_object_path(resource))
+      click_link('Delete Generic')
+      expect(page).to have_content('Work to edit as been deleted')
```
cam156: @awead can you also check that the object is gone here?

awead:
@cam156 fixed.

PR #291, opened 2017-10-18
Custom FindUsing query

awead:
Allows you to query for any attribute in a Valkyrie-backed resource.


awead:
@cam156 I know you're busy with Fedora stuff, so this can wait, but I just wanted to give a little background. I chose not to store MAP fields in Solr for now, and just persist them to Postgres. We can revisit this decision if you like. But, that leaves us with a need for ad-hoc queries which Valkyrie doesn't support (on purpose).

This PR enables us to query our Valkyrie models for a particular value in a field. I'm using it specifically for finding a MAP field with a give label, but it could have other uses too.

app/valkyrie/find_using.rb
```diff
@@ -0,0 +1,26 @@
...
+  attr_reader :query_service
+  delegate :resource_factory, to: :query_service
+  delegate :orm_class, to: :resource_factory
+
+  def initialize(query_service:)
+    @query_service = query_service
+  end
+
+  def find_using(query)
+    sql = "SELECT * FROM orm_resources WHERE metadata @> '{\"#{query.keys.first}\":\"#{query.values.first}\"}';"
```
cam156: @awead is there a reason you are only using first?  Maybe the parameters should just be key, value if you are not intending to support multiple keys and value pairs.

awead: @cam156 `query` is a hash, but should only have one pair. However, we could expand this to be an AND/OR query as well which would handle multiple k/v pairs. I wanted the query to look like our other queries in AR and AF, namely something like: `find_using(name: "Adam")` and not `find_using("name", "Adam")`. I add an exception if `query` has more than one pair of keys and values.

cam156: @awead maybe just a comment for now?

awead:
@cam156 I went with raising an exception.

PR #292, opened 2017-10-19
Adding display transformation base

cam156:
Here is another way to do the Factory.

Not sure if this better or worse than the validation pattern.  The is a single factory for the system, which can have the transformations set on it.

### DIFF
--- a/app/cho/display_transformation/factory.rb
+++ b/app/cho/display_transformation/factory.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+module DisplayTransformation
+  class Factory
+    def self.transformations=(new_transformations)
+      non_transformations = new_transformations.reject { |_label, transformation| transformation.is_a? DisplayTransformation::Base }
+      if non_transformations.present?
+        error_list = non_transformations.keys.join(', ')
+        raise Error.new("Invalid display transformations(s) in transformation list: #{error_list}")
+      end
+
+      @transformations = new_transformations.merge(none: DisplayTransformation::None.new)
+    end
+
+    def self.transformations
+      @transformations ||= { none: DisplayTransformation::None.new }
+    end
+
+    def self.lookup(transformation_name)
+      transformations[transformation_name]
+    end
+  end
+end
--- a/spec/cho/display_transformation/factory_spec.rb
+++ b/spec/cho/display_transformation/factory_spec.rb
@@ -1,0 +1,71 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe DisplayTransformation::Factory, type: :model do
+  before (:all) do
+    class MyTransformation < DisplayTransformation::Base
+    end
+    class OtherTransformation < DisplayTransformation::Base
+    end
+  end
+
+  after (:all) do
+    ActiveSupport::Dependencies.remove_constant('MyTransformation')
+    ActiveSupport::Dependencies.remove_constant('OtherTransformation')
+  end
+
+  subject(:factory) { described_class.new }
+
+  it { within_block_is_expected.not_to raise_exception }
+
+  describe '#tranformations=' do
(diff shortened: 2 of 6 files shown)


app/cho/display_transformation/factory.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module DisplayTransformation
+  class Factory
+    def self.transformations=(new_transformations)
+      non_transformations = new_transformations.reject { |_label, transformation| transformation.is_a? DisplayTransformation::Transformation }
```
awead: Can `_label` be removed?

app/cho/display_transformation/factory.rb
```diff
@@ -0,0 +1,23 @@
...
+      if !non_transformations.empty?
```
awead: `non_transformations.present?` reads better and should work with empty arrays.

app/cho/display_transformation/factory.rb
```diff
@@ -0,0 +1,23 @@
...
+        error_list = non_transformations.map { |label, _err_validator| label }.join(', ')
```
awead: Can `_err_validator` be removed here?

app/cho/display_transformation/factory.rb
```diff
@@ -0,0 +1,23 @@
...
+        raise "Invalid display transofrmations(s) in transformation list: #{error_list}"
+      end
+
+      @transformations = new_transformations
```
awead: I like the fact that that the setter is at the end of the method which works as an explicit return.

app/cho/display_transformation/factory.rb
```diff
@@ -0,0 +1,23 @@
...
+    end
+
+    def self.transformations
+      @transformations ||= { none: DisplayTransformation::None.new }
```
awead: Here I think this is much better than the Validation class, where it was getting set in parameters. However, we would we have to explicitly add the none transformation to the factory initialization, unless we want to always include none by default. So would something like:
``` ruby
@transformations.merge(none: DisplayTransformation::None.new)
```
 be better so there's always `none` there by default?

app/cho/display_transformation/transformation.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module DisplayTransformation
+  class Transformation
```
awead: what about `Base` here? Or is that too overloaded? "Transformation-Transformation" sounds wordy.

app/cho/display_transformation/transformation.rb
```diff
@@ -0,0 +1,9 @@
...
+    def transform(_field)
+      raise 'DisplayTransformation.transform is abstract. Children must implement.'
```
awead: How do you feel about `NotImplementedError`? Although, I did find this: http://chrisstump.online/2016/03/23/stop-abusing-notimplementederror/

spec/cho/display_transformation/factory_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      let(:transformations) { { mine: MyTransformation.new, abc: '123' } }
+
+      it { within_block_is_expected.to raise_exception 'Invalid display transofrmations(s) in transformation list: abc' }
+    end
+  end
+
+  describe '#tranformations' do
+    subject { described_class.transformations }
+
+    let(:transformations) { {} }
```
awead: Shouldn't transformations contain `none`?

app/cho/display_transformation/transformation.rb
awead: Also Abstract, or AbstractTransformation

app/cho/display_transformation/factory.rb
cam156: @awead since this is a hash I do not believe it can be.

PR #300, opened 2017-12-01
Adding controlled vocabulary factory and hooking it into the MAP

cam156:
Followed the pattern from the Display Transformation.

If we like this pattern still I think I should also refactor the Validation to use it and possibly extract a generic Factory (As a next step).

Also do we like the separate folders or should this all be under MAP?


awead:
Linked remaining work to #302

PR #301, opened 2017-12-01
FindModel custom query; group shared queries for resources

awead:
This borrows another Rails-ism that would allow us to do Model.find(id)
type type queries.

Combines the other existing common Rails-type queries such as ::all and
::count into a module that is included in our Valkyrie resources.


config/application.rb
```diff
@@ -15,6 +15,7 @@ class Application < Rails::Application
 
     # Add the work_object folder to avoid creating a WorkObject namespace which is unnecessary at this time.
     config.eager_load_paths << Rails.root.join('app', 'cho', 'work_object')
+    config.eager_load_paths << Rails.root.join('app', 'cho', 'shared')
```
cam156: Why duplicate the line here?  Could you add shared to the line above?

spec/cho/shared/common_queries_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  after(:all) do
+    ActiveSupport::Dependencies.remove_constant('SpecialModel')
+  end
+
+  let!(:common_resource) { Valkyrie.config.metadata_adapter.persister.save(resource: CommonResource.new) }
+
+  describe '::all' do
+    subject { CommonResource.all.map(&:id) }
+
+    it { is_expected.to contain_exactly(common_resource.id) }
```
cam156: I know testing with one resource is most efficient, but this does not seem to test the edge cases.  I would be more confortable with atleast zero and one resource.  Even better  would be zero, one, and some other number.

config/application.rb
awead: @cam156 we could do `config.eager_load_paths =+ [paths]`

awead:
@cam156 this has been updated.

PR #317, opened 2017-12-08
Fixing new to load the form

cam156:
When the item is not a change set you get 'ActionView::Template::Error (undefined method 'errors' for #<MetadataApplicationProfile::Field...'


spec/cho/metadata_application_profile/fields_controller_spec.rb
```diff
@@ -130,6 +130,7 @@
     it 'returns a success response' do
       get :new, params: {}, session: valid_session
       expect(response).to be_success
+      expect(assigns(:metadata_application_profile_field)).to be_a(MetadataApplicationProfile::FieldChangeSet)
```
awead: @cam156 we should probably add this to the other controller tests as well.

Issue #326, opened 2017-12-08
Add additional options to the data dictionary


awead:
Need to add help text to field option in data dictionary
Also add indexing requirements for each field

awead:
Child of #15

PR #336, opened 2017-12-12
adding help text and index type to data dictionary field

cam156:
refs #326

### DIFF
--- a/app/cho/data_dictionary/with_index_type.rb
+++ b/app/cho/data_dictionary/with_index_type.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+module DataDictionary::WithIndexType
+  extend ActiveSupport::Concern
+
+  IndexTypes = Valkyrie::Types::String.enum('facet', 'no_facet')
+
+  included do
+    attribute :index_type, IndexTypes
+  end
+
+  def facet!
+    self.index_type = 'facet'
+  end
+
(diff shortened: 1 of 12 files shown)


app/cho/data_dictionary/with_index_type.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module DataDictionary::WithIndexType
+  extend ActiveSupport::Concern
+
+  IndexTypes = Valkyrie::Types::String.enum('facet', 'none')
```
awead: Is "none" in this case not faceted, or not indexed at all?

cam156: @awead not faceted.  I can change it to that.

awead: @cam156 thanks, that would clear it up!

cam156:
@awead it has been updated

cam156:
@mtribone I figure this is not a final UI, so making it pretty is not really in scope with adding fields.  I'm not even sure this screen will live in the final product.  Did you want to add a ticket and mark all the things wrong with this screen or @awead  can create a ticket to wireframe the screen if we decide it will stay?

awead:
@mtribone I've put in a ticket and slotted it for MVP 1, but we can move it around later if needed.

Issue #333, opened 2017-12-11
title can not be a DataDictionary Field


cam156:
If I add title as a Data Dictionary Field **TWICE** (and then restart the server) 

The catalog controller will not load and gives the error `A index_field with the key title_tesim already exists. (RuntimeError)`

This brings the entire system down hard!

We should validate the uniqueness of the data dictionary keys

cam156:
Any field can not be put into the system twice.

PR #337, opened 2017-12-12
Adding uniqueness validation for data dictionary field label

cam156:
refs #333

### DIFF
--- a/app/cho/data_dictionary/field_change_set.rb
+++ b/app/cho/data_dictionary/field_change_set.rb
@@ -3,7 +3,7 @@
 module DataDictionary
   class FieldChangeSet < Valkyrie::ChangeSet
     property :label, multiple: false, required: true
-    validates :label, presence: true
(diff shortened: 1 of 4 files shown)


app/cho/data_dictionary/field_change_set.rb
```diff
@@ -3,7 +3,7 @@
 module DataDictionary
   class FieldChangeSet < Valkyrie::ChangeSet
     property :label, multiple: false, required: true
-    validates :label, presence: true
+    validates :label, presence: true, 'data_dictionary/unique': true
```
awead: Is `'data_dictionary/unique'` somehow an equivalence to `DataDictionary::UniqueValidator` ?

cam156: @awead that is the Validation way to represent a namespaced validation type.  See the ActiveRecord Validation, which valkyrie uses. presence is also converted to a class name PresenceValidator

PR #338, opened 2017-12-12
Use FactoryGirl::create_for_repository

awead:
This uses a new method for creating factory resources that allows us to
use the data mapper pattern. It was copied from Princeton's Figgy
application, which was in turn taking from an issue in FactoryGirl
itself.

FactoryGirl doesn't support persistence via a data mapper pattern, but
the CreateStategyForRepositoryPattern adds the ability to it.

### DIFF
--- a/spec/factories/field.rb
+++ b/spec/factories/field.rb
@@ -12,10 +12,9 @@ FactoryGirl.define do
     display_transformation 'no_transformation'
     multiple false
     validation 'no_validation'
-  end
 
-  to_create do |resource|
-    change_set = DataDictionary::FieldChangeSet.new(resource)
-    Valkyrie.config.metadata_adapter.persister.save(resource: change_set)
+    to_create do |resource|
+      Valkyrie.config.metadata_adapter.persister.save(resource: resource)
+    end
   end
 end
--- a/spec/factories/work_objects.rb
+++ b/spec/factories/work_objects.rb
@@ -4,5 +4,9 @@ FactoryGirl.define do
   factory :work_object, aliases: [:work] do
     title 'Sample Generic Work'
     work_type 'Generic'
+
+    to_create do |resource|
+      Valkyrie::MetadataAdapter.find(:indexing_persister).persister.save(resource: resource)
+    end
   end
 end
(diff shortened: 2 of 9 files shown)


spec/support/create_for_repository.rb
```diff
@@ -0,0 +1,26 @@
...
+  def association(runner)
+    runner.run
+  end
+
+  def result(evaluation)
+    result = nil
+    evaluation.object.tap do |instance|
+      evaluation.notify(:after_build, instance)
+      evaluation.notify(:before_create, instance)
+      result = evaluation.create(instance)
```
cam156: @awead I realize this is copied code, but do you need to utilize a change set here?

spec/factories/work_objects.rb
```diff
@@ -4,5 +4,9 @@
   factory :work_object, aliases: [:work] do
     title 'Sample Generic Work'
     work_type 'Generic'
+
+    to_create do |resource|
+      Valkyrie::MetadataAdapter.find(:indexing_persister).persister.save(resource: resource)
```
cam156: @awed here you are using indexing_persister and in field we are using metadata_adapter persister.  Is there a reason for the difference?

spec/factories/field.rb
```diff
@@ -12,10 +12,9 @@
     display_transformation 'no_transformation'
     multiple false
     validation 'no_validation'
-  end
 
-  to_create do |resource|
-    change_set = DataDictionary::FieldChangeSet.new(resource)
-    Valkyrie.config.metadata_adapter.persister.save(resource: change_set)
+    to_create do |resource|
+      Valkyrie.config.metadata_adapter.persister.save(resource: resource)
```
cam156: @awead do we not need a change set if the resource is new?

awead: No, only if we wanted to verify the entries.

spec/factories/work_objects.rb
awead: The feature tests for work objects need to be in Solr because we're using Blacklight for presentation. The field doesn't use Blacklight for display, so it's not indexed in Solr.

spec/support/create_for_repository.rb
awead: No, not unless we wanted to validate.

Issue #240, opened 2017-08-31
Create a work type template model


cam156:
The template needs to have:

* The existing work type this template is refining/modifying
* An ordered list of applicable fields
* Options available on each field:  
    * singular or multiple
    * required, recommended, or optional
    * default value
    * autocomplete
    * validate - Date
    * other options under # 3 in https://psu.app.box.com/notes/254001359476

By default, all of the above options will be filled in from the data dictionary.

Child of #320

PR #340, opened 2017-12-14
removing DataDictionary::ChangeSetPersister

cam156:
It was a copy of the valkyrie change set persister with better error handeling, which I moved into Valkyrie ChangeSetPersister

refs #240

Additionally renamed the modules DataDictionary::Base and DataDictionary::ChangeSet becuase the code in the module did not match the name.


awead:
@cam156 something seems off with the coverage. -0.08% is fine, but now it's showing -16%.

cam156:
@awead the coverage is at 0.08 off.  I do not see the 16%

awead:
@cam156 ok, that's weird. It was definitely -16% about an hour ago, but that was before I restarted the failed integration build, so that must have been it.

PR #341, opened 2017-12-14
WIP: Collection model (one class)

awead:
This is the collection model, using one class and a property to denote type (library, archival, curated). It follows the exact same pattern with work objects, although the work object model may be incorrect at this point.

It's a bit easier to manage than the other three-class pattern, in my opinion.

I still need to add additional fields from the wireframes, but this is the basic idea.


awead:
@cam156 I'm guessing it'll involve keying off of `collection_type` which would put a lot of logic in the presenter layer.

PR #342, opened 2017-12-14
Collection models

awead:
This is the collection model with three different Valkyrie resources for each collection type (library, archival, curated).


app/cho/collection/controller.rb
```diff
@@ -0,0 +1,67 @@
...
+      @collection = Collection.new
+    end
+
+    # GET /collections/1/edit
+    def edit; end
+
+    # POST /collections
+    # POST /collections.json
+    def create
+      @collection = Collection.new(collection_params)
```
cam156: @awead is this the spot you are talking about in the comments?  Might be a good place for a CollectionFactory?

awead: @cam156 yeah, that's what I was thinking too. It was the `new` action that gave me pause because we wouldn't know which collection type to create without a parameter. We could pass a parameter in the url: `/collection/new?type=archival`

mtribone:
A solution to the 3 collection types and the additional step would be to have a menu item that says “Create Collection” with 3 choices in the dropdown. And those wireframes aren't from the context of the system as a whole, unfortunately, so they can changed as needed.

As you've pointed out, the advantage of the 3 class solution is being able to restrict or extend the classes as those collection types get a little more defined.

awead:
@mtribone @cam156 here's an updated PR, still not complete, but more fleshed-out. There are separate controllers for each collection, which gives us separate views. It's definitely _boring_.

app/cho/collection/controller_behaviors.rb
```diff
@@ -0,0 +1,78 @@
...
+
+module Collection::ControllerBehaviors
+  extend ActiveSupport::Concern
+
+  delegate :metadata_adapter, :storage_adapter, to: :change_set_persister
+  delegate :persister, :query_service, to: :metadata_adapter
+
+  # GET /collections/new
+  def new
+    @collection = change_set_class.new(resource_class.new).prepopulate!
```
cam156: This code is similar to all the valkyrie controllers.  I wonder if this could be at a higher level, aka ControllerBehaviors.

app/cho/collection/common_fields.rb
```diff
@@ -0,0 +1,17 @@
...
+  Visibility = ['public', 'authenticated', 'private'].freeze
+
+  included do
+    attribute :id, Valkyrie::Types::ID.optional
+    attribute :title, Valkyrie::Types::String
+    attribute :subtitle, Valkyrie::Types::String
+    attribute :description, Valkyrie::Types::String
+    attribute :workflow, Valkyrie::Types::Set.member(Valkyrie::Types::String.enum(*Workflow))
+    attribute :visibility, Valkyrie::Types::Set.member(Valkyrie::Types::String.enum(*Visibility))
+  end
```
cam156: Should members be here?

awead: @cam156 I just remembered that the membership information should be in the work object instead of the collection for performance reasons.

cam156:
@awead do we have a ticket for pointing works at collections?  I just want to be certain the fact that works and collections are not connected is documented so we do not believe it is complete.

awead:
@cam156 how's this? #362

Issue #204, opened 2017-07-24
Collection Landing Pages in CHO


ntallman:
From Henry P.
As a curator of a specific CHO collection, 
I want some control over the public presentation of that collection, 
so that I can market its use, describe its value, credit contributors, acknowledge donors, and similar functions that are normal responsibilities of a curator.

From Ben G.
If splash pages are supported in CHO Access, we want there to be Search, browse, rotating representative images, link to finding aids, But we also prefer the idea of 'finding aids' (in an improved form) serving as de facto splash pages

ntallman:
Working on additional information as to what needs to be contained on a collection landing page.

ntallman:
#583 duplicates this ticket, closing this one.

PR #344, opened 2017-12-15
moving WorkObject into a namespace

cam156:
This allows for additinal WorkObject related items to not need the workObject prefix in thier class name

refs #204  Doing this so I can add a Template to the WorkObject set of things.


app/cho/data_dictionary/fields_for_change_set.rb
```diff
@@ -6,9 +6,14 @@ module DataDictionary::FieldsForChangeSet
   extend ActiveSupport::Concern
 
   included do
-    DataDictionary::Field.all.each do |field|
-      property field.label.parameterize.underscore.to_sym, multiple: field.multiple?, required: field.required_to_publish?
-      validates field.label.parameterize.underscore.to_sym, presence: field.required_to_publish?
+    # There are some raise conditions where the database does not yet exists, but the class is being loaded.
```
awead: Do you mean: "There are some race conditions where the database does not yet exist" ?

app/cho/data_dictionary/fields_for_object.rb
```diff
@@ -6,8 +6,13 @@ module DataDictionary::FieldsForObject
   extend ActiveSupport::Concern
 
   included do
-    DataDictionary::Field.all.each do |field|
-      attribute field.label.parameterize.underscore.to_sym, Valkyrie::Types::Set
+    # There are some raise conditions where the database does not yet exists, but the class is being loaded.
```
awead: Same here: "There are some race conditions where the database does not yet exist" ?

config/application.rb
```diff
@@ -13,7 +13,7 @@ class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
     config.load_defaults 5.1
 
-    # Classes such as WorkObject and shared modules are not namespaced and must be loaded explicitly.
+    # Classes such in shared are not namespaced and must be loaded explicitly.
```
awead: Correct to: "Classes in the shared directory are not namespaced and must be loaded explicitly"

config/application.rb
```diff
@@ -13,7 +13,7 @@ class Application < Rails::Application
...
     config.eager_load_paths += [
       Rails.root.join('app', 'cho', 'work_object'),
```
awead: Can we remove this line now?

spec/cho/work_object/edit_spec.rb
```diff
@@ -9,7 +9,7 @@
   context 'with all the required metadata' do
     it 'updates an existing work with new metadata' do
       visit(edit_work_object_path(resource))
-      fill_in('work_object[title]', with: 'Updated Work Title')
+      fill_in('work_object_object[title]', with: 'Updated Work Title')
```
awead: Is there a way we can not have `work_object_object`? What about just calling the module `Work` instead of `WorkObject`?

spec/support/seed_map.rb
```diff
@@ -4,7 +4,12 @@
 class SeedMAP
   class << self
     def create
-      Valkyrie.config.metadata_adapter.persister.save(resource: title_field)
+      # There are some raise conditions where the database does not yet exists, but the class is being loaded.
```
awead: Needs the same correction here.

spec/rails_helper.rb
```diff
@@ -22,10 +22,6 @@
 require 'rspec/rails'
 # Add additional requires below this line. Rails is not loaded until this point!
 
-# Requires supporting ruby files with custom matchers and macros, etc,
-# in spec/support/ and its subdirectories.
-Dir[Rails.root.join('spec', 'support', '**', '*.rb')].each { |f| require f }
-
```
awead: Is this so we can load seed_map explicitly? How will the other files under the support directory get loaded?

cam156: @awead it was a duplicate.  There is still a requires at line 38

PR #346, opened 2017-12-18
Removing subject and author from catalog controller

cam156:
If the user wants a subject or author field they should define a DataDictionary field.  This refrence conflicted with the data dictionary fields


cam156:
@awead  I deployed this branch to QA to fix the issue we saw with #333.  What actually happened was that having subject as an item conflicted with having subject hardcoded in the catalog controller.  Title is returning the error message, just like every field.

PR #347, opened 2017-12-19
Adding a metadata Schema that can be used for Works or Collections

cam156:
Needed to update valkyrie to get my fix for inheritance

This is a second PR that should be rebase off of master after #344 has been merge.  (The first commit in this PR is the commit for #344).

This PR creates two models in the system a TemplateField, and a Template.  The assumption is the Template field contains most of the information and the Template basically contains a list of those fields.

I am not certain I am in love with storing the TemplateField ids within the template, but I did not want to store them in two places (inside the model, and in the database as their own entity).  I added helper methods so the end user does not need to translate between model and ids.


awead:
@cam156 I had held off until we had addressed everything in #344. But, I should look at this in light of yesterday's discussions.

app/cho/work_object/template.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+module WorkObject
+  class Template < Valkyrie::Resource
+    include CommonQueries
+
+    attribute :id, Valkyrie::Types::ID.optional
+    attribute :label, Valkyrie::Types::String
+    attribute :core_fields, Valkyrie::Types::Array
+    attribute :fields, Valkyrie::Types::Array
```
awead: Arrays are ordered in Postgres. See https://github.com/samvera-labs/valkyrie/issues/301. Would we still need the order index in `TemplateField` or can this just be a set of ordered field ids?

app/cho/work_object/template.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+module WorkObject
+  class Template < Valkyrie::Resource
```
awead: @cam156 based on yesterday's kickoff, I'm wondering if this is now an actual WorkType and not a template. In other words, is this one of the base types, such as document, still image, etc. Would `WorkType::Base` and `WorkType::Field` make more sense? They seem to be their own things, and separate from work objects, even though they will have relationships to them.

app/cho/work_object/template.rb
cam156: @awead yes we need order, since the template will include the order of the fields.

cam156: @awead I guess the question is, if we want to be open to other back ends should we depend on a quirk of one?  I'm ok either way...

app/cho/work_object/template.rb
cam156: @awead The question I ask (and I do not know the answer) is are the Work Types an object with a template, or an instance of a template.  Aka are there other things that the type will need like storage for the actual files, and other things like that.

app/cho/work_object/template.rb
awead: @cam156 that's a good point, and it's nice to be explicit. However, when it comes to other things that we want ordered, like metadata fields, it's awfully nice just to know that what comes in and out of the postgres adapter is in the same order. Let's leave this as-is for now, with the sorting integer. We can revisit this later, if needed.

app/cho/work_object/template.rb
awead: As I understand it, and we may need clarification, work types are all instances. There are 6 "base types" and later we'll have the ability to define sub-types that inherit from these base types. This to me implies that you'd define all the base and sub-types as individual database resources, but they'd all be instances of WorkType. I didn't think storage was a concern for work types, it was only ever a set of metadata fields with their different customized settings. Again, I think we need more clarification here.

cam156:
@awead I renamed the classes based on the discussion yesterday, but now I am wondering if this should be in its own folder or in shared since it will be used by both Works and Collections.

awead:
@cam156 What about a `Schema` namespace? We could put the processing and display schemas in there as well. So you'd have classes like `Schema::Metadata`, `Schema::MetadataField`, `Schema::Display`, `Schema::Processing`. As always, naming is hard.

app/cho/schema/metadata.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    attribute :id, Valkyrie::Types::ID.optional
+    attribute :label, Valkyrie::Types::String
+    attribute :core_fields, Valkyrie::Types::Array
+    attribute :fields, Valkyrie::Types::Array
+
+    alias core_field_ids core_fields
+    alias field_ids fields
+
+    # todo how are we defining the core fields?
```
awead: @cam156 In the model, I had added a boolean property to the data dictionary that would indicate if a field in the dictionary was a core field. That was based on some feedback from Nathan on the original model diagram. See #363

Issue #348, opened 2018-01-29
Implement set of required work types


awead:
Six work types are required: document, still image, audio, moving image, map, and generic.

Each type will have properties, outlined in the wireframe, however, no UI should be present at this time. The UI elements may become the subject of a later release.

awead:
Subtypes will be a later requirement that would affect the implementation of this core 6 types. See #327

awead:
Related to #240

PR #368, opened 2018-02-05
Renaming WorkObject to Work

awead:
After our meeting last week, work objects are now called works. This relates to #348 because I think having `Work` as a module namespace makes sense for classes such as `Work::Deposit` and `Work::Type`


app/cho/work/change_set.rb
```diff
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
-module WorkObject
+module Work
   class ChangeSet < Valkyrie::ChangeSet
-    # self.fields = WorkObject::WorkObject.fields
+    # self.fields = Work::Work.fields
```
cam156: @awead I know this is not your comment, but this does make sense.  Any chance you know what it was supposed to be or maybe we should delete it.

app/cho/work/deposits_controller.rb
```diff
@@ -54,17 +54,12 @@ def validate_save_and_render(change_set, error_view)
         end
       end
 
-      # Never trust parameters from the scary internet, only allow the white list through.
```
cam156: 👍    I was looking at deleting this, but could not figure out why it was there in the first place.

app/cho/work/change_set.rb
awead: Yeah, that's definitely not right... It should be Work::Deposit.

app/cho/work/deposits_controller.rb
awead: @cam156 I think it was part of the old generated controller code.

awead:
@cam156 yes, I think it's a good idea to rename deposit as well. Submission makes the most sense to me right now.

PR #369, opened 2018-02-05
Adding a validate and save that can function around a change set

cam156:
@awead here is an idea for encapsulating our desire to capture the exceptions in the persister and always returning a change set

If you like this I can update the other controller to use this pattern too.


awead:
@cam156 I like it. I think it's an overall improvement to the controller.

cam156:
@awead I updated the other controllers.  Let me know what you think.  I also added in the method to use for buffering, but the tests pass with either method being called in the controllers.

PR #370, opened 2018-02-06
Add find_using as a class method

awead:
This takes the existing find_using custom query and extends it to
Valkyrie resources so that we can run ad hoc queries on specific
resource types. Ex:

    >  MyClass.find_using(label: 'foo')
    => [MyClass< label: foo>]

I've found myself needing this feature lately as I look for specific work types and fields via their labels. It's been aliased to `where` to give it a more ActiveRecord flavor.

### DIFF
--- a/app/cho/shared/common_queries.rb
+++ b/app/cho/shared/common_queries.rb
@@ -13,5 +13,11 @@ module CommonQueries
     def find(id)
       Valkyrie.config.metadata_adapter.query_service.custom_queries.find_model(model: self, id: id)
     end
+
+    def find_using(query)
+      query[:model] = self
+      Valkyrie.config.metadata_adapter.query_service.custom_queries.find_using(query)
+    end
+    alias_method :where, :find_using
   end
 end
--- a/spec/cho/shared/common_queries_spec.rb
+++ b/spec/cho/shared/common_queries_spec.rb
@@ -7,6 +7,7 @@ RSpec.describe CommonQueries do
     class CommonResource < Valkyrie::Resource
       include CommonQueries
       attribute :id, Valkyrie::Types::ID.optional
+      attribute :label, Valkyrie::Types::String
     end
   end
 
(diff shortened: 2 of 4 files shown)


spec/valkyrie/find_using_spec.rb
```diff
@@ -41,4 +41,19 @@ class SampleResource < Valkyrie::Resource
       }.to raise_error(ArgumentError, 'only one query term is supported')
     end
   end
+
+  context 'when providing a model' do
+    subject { saved_resource }
+
+    let(:sample_resource) { SampleResource.new(framjam: 'testing with model') }
```
cam156: @awead It would be good to create a second resource here, so that in the future we do not break something and just return the first of only record.

spec/cho/shared/common_queries_spec.rb
```diff
@@ -64,4 +65,19 @@ class CommonResource < Valkyrie::Resource
...
+    let(:resource2) { CommonResource.new(label: 'second resource') }
+
+    before do
+      Valkyrie.config.metadata_adapter.persister.save(resource: resource1)
+      Valkyrie.config.metadata_adapter.persister.save(resource: resource2)
+    end
+
+    it 'retrieves a resource based on its label' do
+      retrieved_resource = CommonResource.find_using(label: 'first resource')
+      expect(retrieved_resource.first.label).to eq('first resource')
```
cam156: Is there an id you could check here?  That would be a tighter test in my mind.

app/cho/shared/common_queries.rb
```diff
@@ -13,5 +13,11 @@ def all
     def find(id)
       Valkyrie.config.metadata_adapter.query_service.custom_queries.find_model(model: self, id: id)
     end
+
+    def find_using(query)
+      query[:model] = self
+      Valkyrie.config.metadata_adapter.query_service.custom_queries.find_using(query)
+    end
+    alias_method :where, :find_using
```
cam156: Is there a reason for the alias?  I would recommend a test if it is needed.

awead: Calling `.where` might feel more familiar, coming from ActiveRecord. I can go either way...

awead:
@cam156 fixed up. I created a different resource class with the same attribute, which actually showed that the query was not working correctly, so I've updated the sql to fix that.

PR #371, opened 2018-02-06
Truncate orm_resources table in test

awead:
This prevents the test suite from failing to start after debugging a
previous run. If you use `byebug` and then forcibly quit the debugger,
there are artifacts leftover in the database that must be removed even
before the application can start up.

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -1,5 +1,20 @@
 # frozen_string_literal: true
 
+# Remove any artifacts leftover in the database from a previous test run.
+#   Because fields are assigned dynamically, this must be done here before
+#   the test suite is initialized.
+conn = PG.connect(dbname: Rails.configuration.database_configuration['test']['database'])
+
+sql = "SELECT EXISTS (
+   SELECT 1
+   FROM   information_schema.tables
+   WHERE  table_schema = 'schema_name'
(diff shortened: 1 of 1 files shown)


config/environments/test.rb
```diff
@@ -1,5 +1,20 @@
...
+
+sql = "SELECT EXISTS (
+   SELECT 1
+   FROM   information_schema.tables
+   WHERE  table_schema = 'schema_name'
+   AND    table_name = 'orm_resources'
+   );"
+
+exists = conn.exec(sql).values.reduce
+conn.exec('truncate table orm_resources') unless exists.include?('f')
```
cam156: @awead I wonder if it would be simple to just truncate?  I'm fine with this though...

awead: @cam156 I was getting a "table doesn't exist" error in Travis if the table hadn't been created yet. I'm open to finding more elegant solutions in the future. This does seem a bit hacky.

cam156: @awead I believe the correct sql would be `truncate table orm_resources if exists orm_resources`

cam156: @awead  never mind postgres does not support that.  GUess we need to go with hacky for now...

PR #372, opened 2018-02-08
 Implement set of required work types

awead:
This gives us the ability to create new works with pre-defined work types. It does not include the changes required to make the work form display only fields that are available via the work type. I'd like to see that work get separated out into a different ticket, because there may be some outstanding questions, but mainly I wanted to get these changes up and in front of our eyes sooner rather than later.

works

### DIFF
--- a/app/views/work/submissions/new.html.erb
+++ b/app/views/work/submissions/new.html.erb
@@ -1,24 +1,31 @@
-<h1>New Work</h1>
+<%# move this into the change set or some other kind of form object %>
+<% work_type = Work::Type.find(Valkyrie::ID.new(@work.work_type)) %>
 
-<%= form_for(@work.model, url: works_path) do |form| %>
+<% if work_type %>
+  <h1>New <%= work_type.label %> Submission</h1>
 
-  <%- @work.fields.except('work_type').each do |field_name, field_value| %>
-    <p>
-      <%= form.label field_name %><br>
-      <%= form.text_field field_name %>
-    </p>
-  <%- end  %>
+  <%= form_for(@work.model, url: works_path) do |form| %>
 
-  <p>
-    <%= form.label :work_type %><br>
-    <%= form.select :work_type, ["Generic", "Still Image", "Map", "Document", "Audio/Visual"] %>
-  </p>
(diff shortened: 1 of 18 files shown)


app/views/work/submissions/new.html.erb
```diff
@@ -1,24 +1,31 @@
-<h1>New Work</h1>
+<%# move this into the change set or some other kind of form object %>
+<% work_type = Work::Type.find(Valkyrie::ID.new(@work.work_type)) %>
```
cam156: wouldn't this just be @work.label_for_id

app/views/work/submissions/new.html.erb
```diff
@@ -1,24 +1,31 @@
...
 
-<%= form_for(@work.model, url: works_path) do |form| %>
+<% if work_type %>
+  <h1>New <%= work_type.label %> Submission</h1>
```
cam156: @awead Didn't we say that this should stay Work in the UI?

spec/cho/home_page_spec.rb
```diff
@@ -19,5 +16,10 @@
     expect(page).to have_content('New Library Collection')
     click_link('Curated Collection')
     expect(page).to have_content('New Curated Collection')
+    expect(page).to have_link('Create Submission')
+    click_link('Generic')
+    expect(page).to have_content('New Generic Submission')
+    click_link('Document')
```
cam156: @awead does the menu stay open when you click on Generic?  This seems like it may be clicking something else

app/views/work/submissions/new.html.erb
awead: @cam156 I think that's a Rails meta-code thing. I don't have a method with that name.

spec/cho/home_page_spec.rb
awead: @cam156 because we're not using JS, the links are all visible, so we can click on them. It's cheating a bit, but it is testing the links.

app/views/work/submissions/new.html.erb
awead: This will be fixed in #377

PR #380, opened 2018-02-12
PR template

awead:
This is a template that would we used each time we open an new PR. It
should help us to remember what we need to cover and remind us to make
any new documentation if needed.

### DIFF
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,0 +1,15 @@
+Fixes # .
+
+## Description
+
+Why was this necessary?
+
+## Changes
+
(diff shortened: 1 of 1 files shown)


pull_request_template.md
```diff
@@ -0,0 +1,15 @@
+Fixes # .
```
mtribone: Will there be times when something refs instead of fixes?

awead: @mtribone sure. This can be edited as needed. It's just a reminded to link the PR to a relevant ticket.

Issue #376, opened 2018-02-09
Work types should control metadata entry


awead:
The work type needs to dictate which fields are available for editing a new work.

awead:
The actual fields used for each type will be provided later. For now, we can use an example field for each type to demonstrate that the process works.

Issue #322, opened 2017-12-08
Define core set of fields


awead:
As a metadata librarian
I want to specify a set of core fields that are available to every object (both works and collections)
So that we have a minimum level of description across the repository

ntallman:
For MVP, collections will only get the coreMAP fields. Future enhancement will be to add additional fields from data dictionary.

awead:
We'll need a list of what the fields are from @ruthtillman and @lmballinger. After which the fields will be defined in code.

ntallman:
Ruth and Linda will try to get the coreMAP to the dev team by Jan 29, but it might be latter. Moving this story to a lower priority within the sprint to give them some more time, if needed.

awead:
Once we have the list, we'll need to decide how to implement this. Currently, the data model has a boolean field in the data dictionary to indicate if a field is core or not. Are there are other way we need to consider?

awead:
Use CSV import process in SeedMAP

PR #388, opened 2018-02-14
Adding core field flag to data dictionary

cam156:
refs #322.

## Description

Adding a flag to the data dictionary so that core fields can be define in the data dictionary.  I need this for #376 becuase the metadata schema needs to be able to get both core and additional fields from the data dictionary and we need a way to split them up.

## Changes

1. Adding a Flag to the Data Dictionary
2. Updating Tests to use the flag

## Checklist

- [x] i18n enabled
- [x] adequate test coverage
- [x] accessible interface


awead:
@cam156 I've approved this. If you can rebase, go ahead and merge it.

PR #395, opened 2018-02-16
 creating a metadata schema for each work type in seed

cam156:
## Description

References ticket: #376

Why was this necessary?  To be able to show different work types in the form we need a valid schema for each work type with fields that specific to that work type.

## Changes

  1. Changing seed to use csv load
  2. changing tests to use db:seed instead of ther own process
  3. adding a service for generating Metadata::SchemaFields based on the core fields
  4. adding a method for querying core fields to the data dictionary

## Checklist

- [x] i18n enabled
- [x] adequate test coverage
- [x] accessible interface


spec/support/database_cleaner.rb
```diff
@@ -3,7 +3,7 @@
 RSpec.configure do |config|
   config.before(:suite) do
     DatabaseCleaner.clean_with(:deletion)
-    SeedMAP.create
+    Rails.application.load_seed # loading seeds
```
awead: Is `spec/support/seed_map.rb` even needed anymore if you're doing it here?

spec/cho/schema/metadata_core_fields_spec.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require 'valkyrie/specs/shared_specs'
```
awead: Is this needed? I don't see any shared specs in the test.

app/cho/schema/metadata_core_fields.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class Schema::MetadataCoreFields
```
awead: Is this a class or a service? Maybe CoreFieldsGenerator? Service? dunno... We can also leave it for later.

spec/support/database_cleaner.rb
cam156: @awead yes.  The original code called it once when the seed map was set up and once after the database cleaner ran.  I believe the call on load sets up the database for code creation, and the call here sets it up for the test.  I'll add some comments

spec/cho/schema/metadata_core_fields_spec.rb
cam156: @awead good point.  Copy/Paste

app/cho/schema/metadata_core_fields.rb
cam156: @awead I thought it was a generator (since I put the main method as generate.  I could rename it if you like.

awead: @cam156 nah, let's wait and see if a pattern emerges later.

PR #397, opened 2018-02-19
Using Work::Form to change the form based on type

cam156:
## Description

References ticket: #376 

This allows the form to display differently based on the work type.  Each Work Type includes a metadata schema, which defines the fields and their order for display.  This PR modifies the code to use that data.

## Changes

1. Adds Work::Form class
1. Create Form Class in Controller
1. Modify views to use Form

## Checklist

- [ ] i18n enabled
- [ ] adequate test coverage
- [ ] accessible interface


app/cho/work/form.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  def initialize(change_set)
+    @change_set = change_set
+  end
+
+  def fields
+    return @fields if @fields.present?
+    field_ids = metadata_schema.core_fields + metadata_schema.fields
+    unordered_fields = field_ids.map { |id| Schema::MetadataField.find(id) }
+    @fields = unordered_fields.sort_by(&:order_index)
```
awead: @cam156 can a core field and a field from the schema have the same order index?

app/cho/work/type.rb
```diff
@@ -10,5 +10,9 @@ class Type < Valkyrie::Resource
     attribute :metadata_schema_id, Valkyrie::Types::ID.optional
     attribute :processing_schema, Valkyrie::Types::String
     attribute :display_schema, Valkyrie::Types::String
+
+    def metadata_schema
+      Schema::Metadata.find(metadata_schema_id)
```
awead: Should this be memoized?

app/views/work/submissions/new.html.erb
```diff
@@ -1,15 +1,14 @@
-<%# move this into the change set or some other kind of form object %>
-<% work_type = Work::Type.find(Valkyrie::ID.new(@work.work_type)) %>
+<% work_type = @work.work_type %>
```
awead: Since this is now memoized in the form, do we need to keep the variable?

app/views/work/submissions/new.html.erb
```diff
@@ -1,15 +1,14 @@
...
 
   <%= form_for(@work.model, url: works_path) do |form| %>
 
-    <%- @work.fields.except('work_type').each do |field_name, field_value| %>
+    <%- @work.fields.each do |metadata_field| %>
       <p>
-        <%= form.label field_name %><br>
-        <%= form.text_field field_name %>
+        <%= form.label metadata_field.label, metadata_field.display_name %><br>
+        <%= form.text_field metadata_field.label %>
```
awead: These look identical to lines 6-9 in the edit view. Is it worth it to extract them into a shared partial?

app/cho/work/form.rb
cam156: @awead yes, they can be mixed, I believe.  Aka a specific field could be more important than a core field for a work type.

app/views/work/submissions/new.html.erb
cam156: @awead I was thinking the form was exactly the same from the get go, but was not sure why there were two copies.  What was the original reason for two?

awead: @cam156 I can't recall that there was.

cam156:
@awead I think this is all fixed up.  Let me know if you  would like additional changes.

Issue #359, opened 2018-01-30
Upload File(s) to a Single Work Object


ntallman:
As a digital production technician,
I would like to upload a file or files to a work object,
So that the content may be ingested to CHO and made available online.

awead:
Depends on #364

awead:
Includes creating the necessary Valkyrie resources to store the file, as well as any other Ruby classes. FIles will be stored in local temp space for now.

awead:
File upload is not working in QA.

PR #401, opened 2018-02-23
adding in files to the work class

cam156:
## Description

Add in File upload for a work.  This just attaches singular files to a work.  There no components for displaying the contents of the file, downloading the file, or deleting the file.

References ticket: (if any) #359 

Why was this necessary?

## Changes

* added in shrine gem
* added Work::File object
* added file to the Submission::ChangeSet
* added file persistence to the ChangeSetPersister
* added html5 input to the edit and new screen
* added display of attached files to the work show screen

## Checklist

- [x] i18n enabled
- [x] adequate test coverage
- [x] accessible interface


app/cho/work/submission.rb
```diff
@@ -8,5 +8,8 @@ class Submission < Valkyrie::Resource
 
     attribute :id, Valkyrie::Types::ID.optional
     attribute :work_type, Valkyrie::Types::String
+    attribute :files, Valkyrie::Types::Set
+
+    attr_accessor :file_data, :cached_file_data, :file
```
awead: Could you comment more on what these are? I'm assuming they're from Shrine. I'm also confused about `files` versus `file`. The attributes is multiple, but there's a `file` accessor... how does that differ?

app/cho/work/submission_change_set.rb
```diff
@@ -4,6 +4,7 @@ module Work
   class SubmissionChangeSet < Valkyrie::ChangeSet
     validates :work_type, presence: true
     property :work_type, multiple: false, required: true
+    property :file, multiple: false, required: false
```
awead: Why is the change set singular while the model is multiple? Are we only uploading one file at a time?

app/cho/work/file.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class Work::File < Valkyrie::Resource
+  include CommonQueries
+
+  attribute :id, Valkyrie::Types::ID.optional
+  attribute :original_filename, Valkyrie::Types::String
+  attribute :use, Valkyrie::Types::Set
+  attribute :file_identifier, Valkyrie::Types::ID.optional
```
awead: Is this another UUID-type identifier for the file, or something different?

app/cho/work/submission_change_set.rb
cam156: @awead yes at this point I only set up shrine to allow for one file

app/cho/work/file.rb
cam156: @awead that is the actual binary content coming out of storage

awead: @cam156 so if I call `resource.file_identifier` I get binary data back? That seems odd given that the field type is an id. If this is a Shrine/Valkyrie thing, that's fine, but what about a better attribute name such as `content` or `file_handle`?

cam156: @awead it is an identifier for the storage object, so you still need to call find.  I took this concept from valkyrie: https://github.com/samvera-labs/valkyrie/blob/master/spec/valkyrie/derivative_service_spec.rb

Issue #373, opened 2018-02-09
Required to save option


awead:
Fields like "title" should be a required field when saving a new resource. This should be distinct from "required to publish" where a field can be absent, but the resource can still be saved.

What done looks like:
1. Add "required" option to list of requirement designations in the data dictionary
2. The "required" option is interpreted by the form to show which fields are required to save the resource
3. If a required field is blank, the error message should be rendered by the client using HTML5. Note: collections are already done this way with respect to "Title" but works are not.

cam156:
This is using required to publish to mark required: https://github.com/psu-libraries/cho/blob/9b7db9cd69e1beac3ed60fd88de9b172851e6a01/app/cho/data_dictionary/fields_for_change_set.rb#L14-L15

It should use the new required option

PR #410, opened 2018-02-27
Title is Required for Works

mtribone:
## Description

Title is required for the creation of works. Loops through the data dictionary and looks for the fields with the required attribute and adds the additional aria attribute and the span for the required text. Replaces `required_to_publish` with `required` where needed.

References ticket: (if any)
#373 

Why was this necessary?

We need at least one identifying field, so that librarians can find unfinished work in a listing of works.

## Changes

Adds aria required attributes for the title input. Adds the requireddesignation to the data dictionary ruby file and the csv.

Updates style sheet to add bottom margin to all inputs and make the submissions form look for a required designation and add CSS and required attributes to the field. Uses the label for the block, makes the lable titleized, and adds the critical required span. Updates some tests for the title field to be required and no longer just required to publish.

Switches required to publish to required in the fields for change set so that the other change set tests pass without modification. We still want to test for the absence of title.

## Checklist

- [x] i18n enabled
- [x] adequate test coverage
- [x] accessible interface

### DIFF
--- a/app/assets/stylesheets/form.scss
+++ b/app/assets/stylesheets/form.scss
@@ -5,7 +5,7 @@
 h1        { margin-bottom: 2rem; }
 fieldset  { margin-bottom: 3rem; }
 legend    { border: none; }
-.basic_metadata input { margin-bottom: 1rem; }
+input     { margin-bottom: 1rem; }
 .no_bold  { font-weight: normal; }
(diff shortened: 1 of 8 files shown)


mtribone:
screen shot 2018-02-27 at 9 13 33 am

Issue #350, opened 2018-01-29
Basic User Role (MVP only)


ntallman:
Some of our MVP user stories rely on a very basic user role structure.

- Admin user -- anyone who is in the specified UMG
- Public user -- anyone not logged into MVP CHO

Some fields (and possibly other things) will be restricted from public users, only visible to admin users.

Users will be authenticated using LDAP credentials and their access determined using UMGs.

Precedes: #102 
Depends on: #379

awead:
For users who are logged in but not allowed access, the response should be 403 instead of 401. Change the allow_admins method in ApplicationController.

PR #411, opened 2018-02-27
Remote authentication and access

awead:
## Description

References ticket: #350 

Incorporates the devise_remote gem into Cho so we can allow specific users access to the application. I'd like to review this a bit more and then squash some commits. The LdapUser class was copied from Scholarsphere and might be able to be moved into another gem such as LdapDisambiguate.

Why was this necessary?

We need to limit who is allowed to use Cho.

## Changes

* Run the devise_remote installer
* Enable remote authentication with Cosign
* Limit access to users in the cho-admin Ldap group

## Checklist

- [ ] i18n enabled
- [ ] adequate test coverage
- [ ] accessible interface


app/application/application_controller.rb
```diff
@@ -3,7 +3,15 @@
 class ApplicationController < ActionController::Base
   # Adds a few additional behaviors into the application controller
   include Blacklight::Controller
+  include DeviseRemote::HttpHeaderAuthenticatableBehavior
+
   layout 'blacklight'
 
   protect_from_forgery with: :exception
+  before_action :authenticate_user!, :allow_admins
```
cam156: @awead is this protecting all pages?

config/travis/hydra-ldap.yml
```diff
@@ -0,0 +1,18 @@
+development:
```
cam156: @awead this should not be put in git.  In Travis we can just ignore ldap, or can you just put fake data since eyou are mocking ldap in the tests?

awead: @cam156 whoops. I meant to use https://github.com/psu-stewardship/scholarsphere/blob/develop/config/sample/hydra-ldap.yml. Hydra::Ldap will complain if there's no config file present.

app/application/application_controller.rb
awead: @cam156 anything that extends from ApplicationController, which is all pages at the moment.

spec/devise/ldap_user_spec.rb
```diff
@@ -0,0 +1,137 @@
+# frozen_string_literal: true
```
cam156: @awead Do we need a ticket for moving this out?

awead: @cam156 #417 We can figure out where it should go.

Issue #381, opened 2018-02-12
Define relationships between works and collections


awead:
Works need to point to collections, for storage, but collections need to know which works they have for display. The first is a persistence issue, the second is a query issue.

This is done when I can add works to collections and have them save, as well as find all the works in a particular collection.

PR #413, opened 2018-02-27
Define relationships between collections and works

awead:
## Description

References ticket: #381

Why was this necessary?

Defines the relationship between works and collections. This is the first of two PRs, the second of which will implement the wireframes that allow the user to assign works to collections. This PR covers persisting the relationship from work to collection, and then retrieving the reverse, collection to works.

## Changes

## Checklist

- [x] i18n enabled
- [x] adequate test coverage
- [x] accessible interface


app/cho/work/submission_change_set.rb
```diff
@@ -5,11 +5,36 @@ class SubmissionChangeSet < Valkyrie::ChangeSet
...
+    validates :member_of, with: :members_exist
+    property :member_of, multiple: true, required: false
 
     include DataDictionary::FieldsForChangeSet
 
     def initialize(*args)
       super(*args)
     end
+
+    def members_exist(field)
```
cam156: @awead what is the return from this method?  True, False, a list of members?

app/cho/work/submission_change_set.rb
```diff
@@ -5,11 +5,36 @@ class SubmissionChangeSet < Valkyrie::ChangeSet
...
+      members = self[field].map { |id| Member.new(id) }
+      return if members.empty?
```
cam156: This should return the return type.

app/cho/work/submission_change_set.rb
```diff
@@ -5,11 +5,36 @@ class SubmissionChangeSet < Valkyrie::ChangeSet
...
+    end
+
+    class Member
+      attr_reader :id
+
+      def initialize(id)
+        @id = Valkyrie::ID.new(id)
+      end
+
+      def exists?
```
cam156: 👍  I wonder if this should be a common query

spec/cho/work/submissions/change_set_spec.rb
```diff
@@ -21,6 +21,10 @@
     it 'has a single work type' do
       expect(change_set).not_to be_multiple(:work_type)
     end
+
+    it 'has multiple parents' do
+      expect(change_set).to be_multiple(:member_of)
```
cam156: @awead I'm not certain how this test is working.  You are not setting multiple parents anywhere.

app/cho/work/submission_change_set.rb
awead: @cam156 it's a list of members, but that doesn't have any effect on the outcome. It's whether or not there are errors after the method is run.

app/cho/work/submission_change_set.rb
awead: I'm not sure I understand. Why doesn't it need to to return the type?

spec/cho/work/submissions/change_set_spec.rb
awead: @cam156 It's testing the assertion that member_of has `multiple: true`. It's tested in the WithMembers test, but I can update it to use two resources instead of just one.

app/cho/work/submission_change_set.rb
cam156: @awead this is returning either nil or an array. It is confusing what theis method does.

spec/cho/work/submissions/change_set_spec.rb
cam156: @awead ok.  I get it.  The test is that is can be multiple not that is has multiple values...

awead: @cam156 it'll make more sense with the UI PR. A work can multiple collections, but only one that is either library or archival, i.e. the home collection. So I'm just verifying that member_of can be multiple here. In the UI, there will be one collection assigned in the new action, and additional curated collections will be assigned as part of another story in MVP2.

app/cho/work/submission_change_set.rb
awead: @cam156 how about returning true or false no matter what? I don't think it really matters what's returned, but consistency is good!

app/cho/work/submission_change_set.rb
cam156: At this point I see 3 options

1.  Add  yard docs to explain what this is doing since it is a public method

1. rename the method to include an bang '!' to indicate that it is modifying the object

    ```
        def members_exist!(field)
          members = self[field].map { |id| Member.new(id) }
          members.each do |member|
            errors.add(field, "#{member.id} does not exist") unless member.exists?
          end
        end
    ```

1. make the method return a boolean (and update errors)
```
    def members_exist?(field)
      members = self[field].map { |id| Member.new(id) }
      all_members_present = true
      members.each do |member|
         unless member.exists?
            errors.add(field, "#{member.id} does not exist") 
            all_members_present = false
         end
      end
      all_members_present
    end
```

app/cho/work/submission_change_set.rb
cam156: Is members an empty array or nil.  If it is an empty array this line could be removed.  Each will do the right thing for an empty array.

app/cho/work/submission_change_set.rb
awead: @cam156 I think both 1 and 2, but maybe rename the method to `validate_membership!` ?

app/cho/work/submission_change_set.rb
awead: @cam156 See #418.

awead:
@cam156 Ok, I think I've got this fixed up. I've side-stepped the `exists?` issue and put in another PR for the common query. One thing that will affect this is the fact that the new `exists?` method would check for a specific class and raise an exception if it found one that didn't match. Here, however, we want to allow multiple types, so that would mean some more logic. We can look at how to do that once that PR is merged.

Issue #383, opened 2018-02-12
Implement collection relationship wireframes


awead:
Implement the wireframes created in #382

awead:
Depends on #381

PR #420, opened 2018-03-01
Implement collection relationship wireframes 

awead:
## Description

Builds off of #413 and adds the UI methods to add works to collections and display them in paginated fashion.

collection

References ticket: #383 

Why was this necessary?

We need the ability to works to collections.

## Changes

This expands and overrides Blacklight a lot in order to display a collection's members in tabular format. I'm not entirely happy with the approach just yet. I borrowed most of it from Figgy and it doesn't have much test coverage yet. I'd like to explore some other options, maybe, before merging this.

No

## Checklist

- [x] i18n enabled
- [x] adequate test coverage
- [ ] accessible interface


app/views/catalog/_document_members.html.erb
```diff
@@ -0,0 +1,19 @@
+<div id="members" class="members-<%= document_index_view_type %>">
+  <table>
+    <thead>
+      <tr>
+        <td><%= t('cho.collection.members.title') %></td>
```
mtribone: An upgrade for accessibility would be to add the row and column attributes to the table.
https://webaim.org/techniques/tables/data#th

PR #423, opened 2018-03-05
WAVE Report Errors and Warnings

mtribone:
Refactors the default view so that the main h1 isn't hidden and isn't repeating the name of the application. Refactors the headings and layout for the search results page, so that we have a good document outline and that the content comes before the facets in the code.

Fixes insuficient color contrast and overrides the layout helper for column sizes.

Adds label for the file upload button and a table header for the data dictionary.

Replaces the search history table with an unorderd list.

Aligns HTML for search history and saved searches.

Rubocop fixes.

Fixes the layout helper from blacklight so that the show page doesn't have the pull class to keep the tools column on the right.

## Description

References ticket: (if any)

Why was this necessary?

## Changes

## Checklist

- [x] i18n enabled
- [x] adequate test coverage
- [x] accessible interface

### DIFF
--- a/app/views/catalog/_search_results.html.erb
+++ b/app/views/catalog/_search_results.html.erb
@@ -1,0 +1,25 @@
+<%# Overrides the view from Blacklight so that we have good heading structure for content %>
+<h1 class="sr-only top-content-title"><%= t('blacklight.search.search_results_header') %></h1>
+
+<% @page_title = t('blacklight.search.page_title.title', :constraints => render_search_to_page_title(params), :application_name => application_name) %>
+
+<% content_for(:head) do -%>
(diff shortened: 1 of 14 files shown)


app/views/catalog/_search_results.html.erb
```diff
@@ -0,0 +1,25 @@
+# @note Overrides the view from Blacklight so that we have good heading structure for content
```
awead: @mtribone It's great that you want to write Yard-style docs! Unfortunately, we can't use it in a view context because Yard doesn't parse those. I think we'll just have to use standard Ruby-style comments with `<%# [comment goes here] %>`

PR #463, opened 2018-04-10
Use a metadata schema for collections

### LINKED ISSUE #407: Create metadata schema for collections
Collection fields are currently defined directly on the model via Collection::CommonFields. Instead, they should be defined using the core fields identified in DataDictionary.

We need a particular instance of Schema::Metadata that's just for collections. This is so the same process of making changes to fields in works can be applied to collections as well.

Depends on #403. The collection change set needs to do the same things that the updated work change set does once that ticket is completed.

Relates to #322

awead:
## Description

Allows us to define the same set of core fields for collections that are
used for works. Additional configuration can later be added to expand
fields on collections and vary them by the collection class.

Fixes #407 

Why was this necessary?

Makes the process for defining fields on works and collections the same.

## Changes

## Checklist

- [x] i18n enabled
- [x] adequate test coverage
- [x] accessible interface

### DIFF
--- a/app/cho/collection/change_set_behaviors.rb
+++ b/app/cho/collection/change_set_behaviors.rb
@@ -2,15 +2,9 @@
 
 module Collection::ChangeSetBehaviors
   extend ActiveSupport::Concern
+  include DataDictionary::FieldsForChangeSet
 
   included do
-    property :title, multiple: false, required: true
-    validates :title, presence: true
-
-    property :subtitle, multiple: false, required: false
-
-    property :description, multiple: false, required: false
-
     property :workflow, multiple: false, required: true
     validates :workflow, inclusion: { in: ::Collection::CommonFields::WORKFLOW }
     validates :workflow, presence: true
@@ -19,4 +13,18 @@ module Collection::ChangeSetBehaviors
     validates :visibility, inclusion: { in: ::Collection::CommonFields::VISIBILITY }
     validates :visibility, presence: true
   end
+
+  # @return [Array<Schema::MetadataField>]
+  def form_fields
+    @form_fields ||= ordered_form_fields
+  end
+
+  private
+
+    def ordered_form_fields
+      metadata_schema = Schema::Metadata.find_using(label: 'Collection').first
+      field_ids = metadata_schema.core_fields + metadata_schema.fields
+      unordered_fields = field_ids.map { |id| Schema::MetadataField.find(id) }
+      unordered_fields.sort_by(&:order_index)
+    end
 end
--- a/app/views/collection/base/_form.html.erb
+++ b/app/views/collection/base/_form.html.erb
@@ -1,14 +1,8 @@
 <%= form_for(@collection.model) do |form| %>
+
   <fieldset class="basic_metadata">
     <legend><%= t('cho.form.metadata') %></legend>
-    <%= form.label :title do %>
-      <%= t('cho.field_label.title') %><span class="required no_bold"> <%= t('cho.field_label.required') %></span>
-    <% end %>
-      <%= form.text_field :title, :required => true, 'aria-required' => true %>
-    <%= form.label :subtitle %>
-      <%= form.text_field :subtitle %>
-    <%= form.label :description %>
-    <%= form.text_area :description %>
+    <%= render 'shared/form_fields', resource: @collection, form: form %>
   </fieldset>
 
   <fieldset>
--- a/config/data_dictionary/schema_fields.yml
+++ b/config/data_dictionary/schema_fields.yml
@@ -1,6 +1,8 @@
 development: &development
   - schema: 'Generic'
     fields: []
+  - schema: 'Collection'
+    fields: []
   - schema: 'Document'
     fields:
       contributor:
@@ -150,4 +152,6 @@ test:
     fields:
       audio_field:
         order_index: 1
+  - schema: 'Collection'
+    fields: []
 production: *development
--- a/spec/cho/collection/change_set_behaviors_spec.rb
+++ b/spec/cho/collection/change_set_behaviors_spec.rb
@@ -28,17 +28,13 @@ RSpec.describe Collection::ChangeSetBehaviors do
 
   describe '#multiple?' do
     it 'has one title and description' do
-      expect(change_set).not_to be_multiple(:title)
-      expect(change_set).not_to be_multiple(:subtitle)
-      expect(change_set).not_to be_multiple(:description)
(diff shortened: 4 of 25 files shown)


app/views/collection/base/_form.html.erb
```diff
@@ -1,14 +1,19 @@
...
     <legend><%= t('cho.form.metadata') %></legend>
-    <%= form.label :title do %>
-      <%= t('cho.field_label.title') %><span class="required no_bold"> <%= t('cho.field_label.required') %></span>
-    <% end %>
-      <%= form.text_field :title, :required => true, 'aria-required' => true %>
-    <%= form.label :subtitle %>
-      <%= form.text_field :subtitle %>
-    <%= form.label :description %>
-    <%= form.text_area :description %>
+    <%- @collection.fields.each do |metadata_field| %>
```
cam156: I wonder if this could be the same partial as the work

config/data_dictionary/schema_fields.yml
```diff
@@ -1,6 +1,8 @@
 development: &development
   - schema: 'Generic'
     fields: []
+  - schema: 'Collection'
```
cam156: Sine this schema is not being utilized properly maybe it should not be created yet.

spec/cho/collection/change_set_behaviors_spec.rb
```diff
@@ -28,27 +29,20 @@ class MyCollectionChangeSet < Valkyrie::ChangeSet
 
   describe '#multiple?' do
     it 'has one title and description' do
-      expect(change_set).not_to be_multiple(:title)
```
cam156: Should this be to be_multiple?

spec/views/archival_collections/edit.html.erb_spec.rb
```diff
@@ -15,7 +15,7 @@
     assert_select 'form[action=?][method=?]', archival_collection_path(@collection), 'post' do
       assert_select 'input[name=?]', 'archival_collection[title]'
       assert_select 'input[name=?]', 'archival_collection[subtitle]'
-      assert_select 'textarea[name=?]', 'archival_collection[description]'
+      assert_select 'input[name=?]', 'archival_collection[description]'
```
cam156: We should add a ticket for this to remain a textarea

app/cho/collection/change_set_behaviors.rb
```diff
@@ -19,4 +13,15 @@ module Collection::ChangeSetBehaviors
     validates :visibility, inclusion: { in: ::Collection::CommonFields::VISIBILITY }
     validates :visibility, presence: true
   end
+
+  # @return [Array<Schema::MetadataField>]
+  # @note For now, all collections only use core fields, but this could be expanded to
+  #   include additional fields, such as administrative metadata, as well as different
+  #   fields for each class of collection.
+  def fields
+    metadata_schema = Schema::Metadata.find_using(label: 'Collection').first
```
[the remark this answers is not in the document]
awead: @cam156 Administrative metadata will eventually go here. Also, this keeps the pattern consistent with works. We can differentiate between core fields used in different works, as well as collections.

app/cho/collection/with_members.rb
```diff
@@ -2,6 +2,8 @@
 
 module Collection
   module WithMembers
+    extend ActiveSupport::Concern
```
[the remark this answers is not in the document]
awead: I think this was leftover from an earlier iteration. I can remove it.

app/views/collection/base/_form.html.erb
awead: It could be a shared partial between the two. I'll add it in and see what it looks like.

config/data_dictionary/schema_fields.yml
awead: The plan would be to put administrative metadata fields here.

spec/cho/collection/change_set_behaviors_spec.rb
awead: @cam156 I was only testing the attributes that were explicitly defined in the module. Title is defined in DataDictionary, which is included in this module, but perhaps tested elsewhere. That said, I'm happy to add a title attribute since it is a required field.

spec/views/archival_collections/edit.html.erb_spec.rb
awead: @cam156 Done. There's already a "text" field type in the data dictionary, so this should be a matter of enabling it.

app/cho/collection/with_members.rb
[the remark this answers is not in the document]
awead: Actually, I can't. It's the `included do` that I need it for, I think.

awead:
@cam156 I've updated the PR. The shared view between work form and collection form feels a bit forced, but I think it's starting to expose some latent refactoring between the two.

app/cho/collection/with_members.rb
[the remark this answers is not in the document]
cam156: There is no included do in this file though...

app/cho/collection/change_set_behaviors.rb
```diff
@@ -19,4 +13,15 @@ module Collection::ChangeSetBehaviors
...
     validates :visibility, presence: true
   end
+
+  # @return [Array<Schema::MetadataField>]
+  # @note For now, all collections only use core fields, but this could be expanded to
+  #   include additional fields, such as administrative metadata, as well as different
+  #   fields for each class of collection.
+  def form_fields
+    metadata_schema = Schema::Metadata.find_using(label: 'Collection').first
+    metadata_schema.core_fields.map do |id|
```
cam156: Should we be Utilizing the code from Work to fully implement the Schema to Collection mapping?  It is only a couple more lines of code. https://github.com/psu-libraries/cho/blob/master/app/cho/work/submission_change_set.rb#L31-L36
Then there would be no need for the comments about this not being fully implemented.

app/cho/collection/with_members.rb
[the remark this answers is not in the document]
awead: wrong file...

awead:
@cam156 round two.

app/cho/collection/change_set_behaviors.rb
```diff
@@ -19,4 +13,18 @@ module Collection::ChangeSetBehaviors
     validates :visibility, inclusion: { in: ::Collection::CommonFields::VISIBILITY }
     validates :visibility, presence: true
   end
+
+  # @return [Array<Schema::MetadataField>]
```
cam156: 👍

PR #476, opened 2018-04-13
Registering CreateForRepository as create

cam156:
This way we do not need to remember to do something differently to get the saved objects as the result from create

## Description
Create for Valkyrie objects was not working as expected (you were getting the un saved resource back.  You needed to remember to use create_for_repository.  Since almost all of our resources are Valkyrie resources this seemed backwards to me.

References ticket: (if any)

Why was this necessary?

## Changes

* CreateForReporitory is registered as create instead of create_for_repository
* A generic to_create for valkyrie index resources was defined in the rails helper
* Tests were changes to use create instead of create_for_repository
* Any factory that used the indexing persister no longer needed the to_create so that was removed
* The User factory now needed a specialized to_create so that was added

## Checklist

- [ ] i18n enabled
- [ ] adequate test coverage
- [ ] accessible interface

### DIFF
--- a/spec/blacklight/catalog_spec.rb
+++ b/spec/blacklight/catalog_spec.rb
@@ -22,7 +22,7 @@ RSpec.describe 'Catalog page', type: :feature do
   context 'with an existing work' do
     let(:query) { 'exceptional' }
 
-    before { create_for_repository(:work, title: 'Some exceptional content') }
+    before { create(:work, title: 'Some exceptional content') }
 
     it 'searches and facets on the item' do
       visit('/catalog')
--- a/spec/blacklight/solr_document_spec.rb
+++ b/spec/blacklight/solr_document_spec.rb
@@ -22,18 +22,18 @@ RSpec.describe SolrDocument, type: :model do
   describe '#files' do
     subject { SolrDocument.new(document).files.first }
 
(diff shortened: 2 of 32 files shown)


spec/rails_helper.rb
```diff
@@ -55,3 +55,9 @@
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
 end
+
+FactoryGirl.define do
+  to_create do |resource|
+    Valkyrie::MetadataAdapter.find(:indexing_persister).persister.save(resource: resource)
+  end
+end
```
awead: @cam156 could this go under `spec/support/factory_girl.rb` instead?

awead: @cam156 this forces all Valkyrie resources to use the indexing persister. We don't (yet, perhaps?) index work types in Solr, we just store them in Postgres. I believe this is also true of data dictionary terms.

cam156: @awead as you can see in the PR you can override the default.  That is done for Data Dictionary.  Work type is working as an indexed field, but it could be overridden like DataDictionary is.

cam156:
@awead I believe I have addressed the issues.  I also rebased with master. This should be good to go.

Issue #448, opened 2018-03-28
Test for files in works


awead:
Child of #203

Write a test that creates 1000 works each with a 50 MB file and captures the performance metrics.

nrm149:
- Will have to make sure there is enough disk space

PR #479, opened 2018-04-13
Benchmark tests for collections and works

awead:
## Description

Fixes: #448 

Takes existing benchmark code from a rake task and moves it into `cho/metrics` where it can be tested.

* creates a new benchmark for creating works with a 50 MB (or other custom sized) file
* uses Faker gem to fill-out metadata fields

## Changes

Are there any major changes in this PR that will change the release process? No, but we could leverage the benchmark tests to run automatically for each release.

## Checklist

- [x] i18n enabled
- [x] adequate test coverage
- [x] accessible interface


app/cho/metrics/collection.rb
```diff
@@ -0,0 +1,100 @@
...
+    def benchmark
+      Benchmark.benchmark("#{I18n.t('cho.metrics.benchmark_heading')}\n", 0, "%u,%y,%t,%r\n") do |bench|
+        (1..length).each do |count|
+          bench.report { work_resource(count) }
+        end
+      end
+    end
+
+    def output
+      @output ||= File.new("tmp/#{self.class.to_s.split(/::/).last.downcase.pluralize}_#{length}.csv", 'w')
```
cam156: `#{self.class.to_s.split(/::/).last.downcase` is the same code as the i18n_key

spec/cho/metrics/work_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    end
+  end
+
+  it 'creates a work with a 1 MB file and benchmarks the results' do
+    metric.file_size = 1
+    metric.run
+    header, results = report.string.split("\n")
+    expect(header).to eq(' User,System,Total,Real')
+    expect(results.split(',').count).to eq(4)
+    expect(work.files).not_to be_empty
```
cam156: Shouldn't this be one file? with a certain type?

app/cho/metrics/collection.rb
awead: @cam156 it's similar, but not exactly the same same. The i18n_key is something like like "cho.metrics.collection", but we can make the file names match.

Issue #238, opened 2017-08-30
Implement CSV Import


cam156:
Run through The Dry run results and create the objects based on the work already done.

Child of #20

nrm149:
Tickets to be created:
1. Should the csv import be able to circumvent the workflow
2. Set path to file to be relative and not absolute
3. Can we use anything in the identifier field that will map back to the valkyrie collection

PR #485, opened 2018-04-19
Full loop from file to repo

cam156:
## Description

Creates works by importing a csv file in the UI, running a dry run, and import.

Running this brings up some issues.
1. How are we getting the files attached to the work?
1. Should this really be a UI element based on question 1?
1. When the same file gets uploaded twice the upload process errors (files are stored on disk with their original file name)

References ticket: (if any) #238 

Why was this necessary?

## Changes

* Adds a Importer for the CSV Dry Run Results
* Adds files to the DRY Run
* Adds UI elements for running the entire process (very basic)

## Checklist

- [x] i18n enabled
- [x] adequate test coverage
- [ ] accessible interface

### DIFF
--- a/app/cho/work/import/csv_controller.rb
+++ b/app/cho/work/import/csv_controller.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+module Work
+  module Import
+    class CsvController < ApplicationController
+      # GET /csv_file/new
+      def new
+        @csv_file = CsvFile.new
+      end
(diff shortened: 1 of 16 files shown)


cam156:
@awead This could be reviewed.  The UI is very basic, Since the wireframes have not been completed they are just a placeholder.  I'm also still not convinced that we should be doing it this way, but it does work, and could be considered a first step.

PR #486, opened 2018-04-19
Checking that the work type id is valid

cam156:
## Description

References ticket: (if any)

Why was this necessary?

Before you could put any value into the work_id and the change set would be valid.

## Changes

## Checklist

- [ ] i18n enabled
- [ ] adequate test coverage
- [ ] accessible interface


app/cho/work/submission_change_set.rb
```diff
@@ -35,7 +40,12 @@ def form_fields
     end
 
     def work_type
-      @work_type ||= Work::Type.find(Valkyrie::ID.new(work_type_id))
+      id = if work_type_id.is_a?(Valkyrie::ID)
+             work_type_id
+           else
+             Valkyrie::ID.new(work_type_id)
+           end
+      @work_type ||= Work::Type.find(id)
```
awead: @cam156 I've been having this same issue with strings vs. Valkyrie::ID as well. Can we just change `.find` to take either? We could also just mark it `todo` and move on; otherwise, I'm fine with this.

cam156: @awead This would get fixed if we upgraded to Valkyrie 1.0, but there are other issues with upgrading.

How about I add a todo that this code can be removed once we use Valkyrie 1.0?

cam156:
@awead added in the todo

Issue #17, opened 2016-02-08
Export Metadata


jennielevineknies:
As a cataloger, 
I would like the ability to export metadata from across the repository in a variety of formats 
So that I can perform a systematic review of content.

Additional user story from Linda B:
As a metadata manager, I need to be able to export metadata as CSV files, so I can manipulate the metadata outside of the repository.

Related: #20

nrm149:
- When I can export meta data in the same format I use for CSV import

nrm149:
- Dependency with #20

nrm149:
- not implemented in the UI

PR #493, opened 2018-04-25
Adding csv export 

cam156:
* Adds .csv format to show for work and collection
 * Adds method name to field
 * Changed field methods defined on SolrDocument to not conflict with existing methods. resource field vs resource method existing on solr document
 * Changes indexer to use new method name

## Description

References ticket: (if any) refs #17

Why was this necessary? Adding export as requested by users

## Changes

## Checklist

- [x] adequate test coverage

### DIFF
--- a/app/blacklight/document/csv.rb
+++ b/app/blacklight/document/csv.rb
@@ -1,0 +1,97 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+# This module provide CSV export for works and collections
+#  @note the collection export is a list of works in CSV format, not the collection metadata
+#
+module Document::Csv
+  include CsvParsing
+
+  def self.extended(document)
+    # Register our exportable formats
+    Document::Csv.register_export_formats(document)
+  end
+
+  def self.register_export_formats(document)
+    document.will_export_as(:csv)
+  end
+
+  # Export the currect solr document in csv format
+  def export_as_csv
+    if collection?
+      export_collection
+    else
+      export_model
+    end
+  end
+
+  # needed to use the solr document as a scope for a search builder
+  def resource
+    self
+  end
+
+  # needed to use the solr document as a scope for a search builder
+  def blacklight_config
+    CatalogController.blacklight_config
+  end
+
+  private
+
+    def export_collection
+      ::CSV.generate do |csv|
+        csv << ['id', 'members_of_collection_id'] + fields.map(&:label)
+        export_members(csv)
+      end
+    end
+
+    def export_model
+      ::CSV.generate { |csv| csv << export_fields(self) }
+    end
+
+    def export_members(csv)
+      page = 1
+      more_members = true
+      while more_members
+        members = document_facade(page).members
+        members.each do |member|
+          csv << export_fields(member)
+        end
+        page += 1
+        more_members = members.count >= rows
+      end
+    end
+
+    def export_fields(solr_document)
+      values = fields.map do |field|
+        to_csv_field(solr_document.send(field.method_name))
+      end
+      [solr_document.id, to_csv_field(solr_document['member_of_collection_ids_ssim'])].concat(values)
+    end
+
+    def fields
+      @fields ||= DataDictionary::Field.all
+    end
+
+    # cleans up ids and converts arrays to a single string
+    def to_csv_field(field_value)
+      return field_value if field_value.blank?
+      remove_id_marker(Array.wrap(field_value)).join(VALUE_SEPARATOR)
+    end
+
+    # removes id- that gets added to the id fields in solr documents
+    def remove_id_marker(field_value)
+      field_value.map { |value| value.sub(/^id-/, '') }
+    end
+
+    def document_facade(current_page = 1)
+      SolrFacade.new(
+        repository: CatalogController.new.repository,
+        query: CollectionMemberSearchBuilder.new(self).page(current_page).rows(rows).query
+      )
+    end
+
+    def rows
+      100
+    end
+end
--- a/app/cho/work/import/csv_reader.rb
+++ b/app/cho/work/import/csv_reader.rb
@@ -5,6 +5,8 @@ require 'csv'
 module Work
   module Import
     class CsvReader
+      include ::CsvParsing
+
       attr_reader :headers, :csv_hashes
 
       delegate :each, :map, to: :csv_hashes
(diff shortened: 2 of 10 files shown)


app/blacklight/document/csv.rb
```diff
@@ -0,0 +1,87 @@
...
+        page += 1
+        more_members = members.count >= rows
+      end
+    end
+
+    def export_model
+      ::CSV.generate { |csv| csv << export_fields(self) }
+    end
+
+    def export_fields(model)
```
awead: what's model here? A solr document or the response from SolrFacade? Some Yard docs might be helpful.

app/blacklight/document/csv.rb
```diff
@@ -0,0 +1,87 @@
...
+        format_field_value(model.send(field.method_name))
+      end
+      [model.id, format_field_value(model['member_of_collection_ids_ssim'])].concat(values)
+    end
+
+    def fields
+      @fields ||= DataDictionary::Field.all
+    end
+
+    def format_field_value(field_value)
```
awead: Why is this method necessary? It looks like it's getting applied to all fields, but only does stuff to fields with ids.

app/cho/work/submission_indexer.rb
```diff
@@ -27,7 +27,7 @@ def label_for_id
 
       def collection_labels_for_ids
         return unless resource.respond_to?(:member_of_collection_ids)
-        resource.member_of_collection_ids.map { |id| SolrDocument.find(id).title }.flatten
+        resource.member_of_collection_ids.map { |id| SolrDocument.find(id).field_title }.flatten
```
awead: I'm not clear on why we have to use `field_title` instead of `title`? Is this now true of all the fields in a SolrDocument? Will we have to append `field_` to all of them if we want to get the values back?

spec/blacklight/solr_document_spec.rb
```diff
@@ -42,10 +42,61 @@ class MyResource; end
...
+      let(:work) { create :work, member_of_collection_ids: [collection.id], title: 'Work One' }
+      let(:work1_csv) { "#{work.id},#{collection.id},Work One,,,,,,,,," }
+      let(:work2) { create :work, member_of_collection_ids: [collection.id], title: 'Work Two' }
+      let(:work2_csv) { "#{work2.id},#{collection.id},Work Two,,,,,,,,," }
+      let(:csv_header) { 'id,members_of_collection_id,title,subtitle,description,created,generic_field,document_field,still_image_field,moving_image_field,map_field,audio_field' }
+      let(:document) { { 'internal_resource_tsim' => 'MyCollection', id: collection.id, title_tesim: ['my_collection'] } }
+
+      before do
+        work
+        work2
```
awead: You should be able to use `let!` here instead of creating them in a before block... your choice...

app/cho/work/import/csv_reader.rb
```diff
@@ -12,8 +12,20 @@ class CsvReader
...
-        @csv_hashes = csv_data.map { |a| Hash[@headers.zip(a)] }
+        @csv_hashes = csv_data.map { |a| Hash[@headers.zip(format_csv_data(a))] }
       end
+
+      private
+
+        def format_csv_data(data_array)
+          data_array.map do |value|
+            if value.present? && value.include?('|')
+              value.split('|')
```
awead: Can we name a constant for "|" something like `Document::Csv::FIELD_SEPARATOR`? Makes it easier to change later, and helps identify it.

app/blacklight/catalog_controller.rb
```diff
@@ -4,6 +4,7 @@ class CatalogController < ApplicationController
   include Blacklight::Catalog
   include CollectionMembers
 
+  blacklight_config
```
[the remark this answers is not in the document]
cam156: No that should have been removed.  I forgot to push...

app/blacklight/document/csv.rb
```diff
@@ -0,0 +1,87 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+# This module provide Dublin Core export based on the document's semantic values
```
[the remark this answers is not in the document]
cam156: copy pasta!  Good catch!

app/blacklight/document/csv.rb
cam156: @awead good point It should be solr_document.  My understanding of a solr_facade, is that the members are solr_documents.  This is a private method used just in this class, would changing the parameter name be enough?

app/blacklight/document/csv.rb
cam156: Yeah, you are correct.  I should rename it clean_ids, or something.  I was thinking it would do more, but that was the only field that needed to be formatted

app/cho/work/submission_indexer.rb
cam156: @awead that was part of the comments on the PR.  We were running into conflicts with Data Dictionary terms and the actual methods on the solr document.

spec/blacklight/solr_document_spec.rb
cam156: @awead rubocop got angry so I changed them from let!  Did we want to change the rubocop rules?

app/blacklight/document/csv.rb
awead: @cam156 I think so. Or adding a doc: `# @param [SolrDocument] document`

spec/blacklight/solr_document_spec.rb
awead: @cam156 nah. always obey the law!

app/cho/work/submission_indexer.rb
awead: @cam156 right, I just didn't understand what the conflict was. Does that mean we have to call `field_x` for everything? 😦

cam156: @awead yes, that was my solution.  field_ could be anything.  I am open to suggestions for a different solution.  I thought about not creating methods if they conflicted, but that seemed less desirable than a constant prefix.  field_x is just for the data dictionary field accessors.  Other methods would remain the same.

awead: @cam156 Ok. As per your suggestion in Slack, can we rename the methods `x_field`, ex. `title_field`? That reads a bit better and folks like me would reach for "title" first in a method name instead of "field". Also, a comment section either in fields_for_solr_document.rb or field.rb that just explains what's going on and that you'll need to use the `_field` suffix name instead of the regular one.

Issue #449, opened 2018-03-28
Implement an accessible interface to delete a single work with a user confirmation


awead:
Child of #119

Deletes a single work and its files and uses an accessible method for confirming the delete process with the user. This could be with RailsJS, or any other method.

nrm149:
- Complete is a delete button on the work edit page; a delete button that works

ntallman:
Child of #119

Issue #450, opened 2018-03-28
Implement an accessible interface to delete multiple works with a user confirmation


awead:
Child of #119

Delete multiple works and their files. Options include an option next to search results in the catalog, followed by a confirmation page, or other solutions.

This ticket may include investigating other options.

nrm149:
- Complete #449 first

ntallman:
Child of  #119

PR #495, opened 2018-04-26
Delete resources via confirmation process

awead:
## Description

Route delete requests to a specific controller that confirms the user's
selection and then deletes the item.

Here's a screencast that shows the whole thing in action:

https://www.dropbox.com/s/vec74mftbdo09fk/batch-delete.m4v?dl=0

Fixes: #449 and #450 

## Changes

Refactors our controllers because deletion is no longer required. Also reworks the ChangeSetPersister to delete files from works and removes unused code.

## Checklist

- [x] i18n enabled
- [x] adequate test coverage
- [x] accessible interface


awead:
There are some edge cases and failure scenarios and are not dealt with in this PR, but ticketed in #497.

app/cho/shared/deletes_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+  def destroy
+    resources.each do |resource|
+      change_set_persister.delete(change_set: DeleteSet.new(resource))
+    end
+    flash[:deleted_titles] = resources.map(&:title).flatten
+    redirect_to(deletes_path)
+  end
+
+  # GET /deletes
+  def index; end
```
cam156: @awead what is this?  Should it be removed?

app/valkyrie/change_set_persister.rb
```diff
@@ -80,37 +69,15 @@ def store_files(change_set)
...
 
-  # def before_delete(change_set:)
-  #   parents = query_service.find_parents(resource: change_set.resource)
-  #   parents.each do |parent|
-  #     parent.member_ids -= [change_set.id]
-  #     persister.save(resource: parent)
-  #   end
-  # end
+    def delete_file(resource:)
+      FileUtils.rm(URI(resource.file_identifier.id).path)
```
cam156: I wonder if this should be moved into a method on file? `resource.path`

app/views/deletes/confirm.html.erb
```diff
@@ -0,0 +1,15 @@
...
+<%= form_for(:delete, url: deletes_path, method: :delete) do |form| %>
+  <ul>
+    <% @resources.each do |resource| %>
+      <li>
+        <%= resource.title.first %>
+        <%= form.hidden_field :ids, multiple: true, value: resource.id %>
+      </li>
+    <% end %>
+  </ul>
+  <%= form.submit t('cho.deletes.confirm.submit') %>
```
cam156: Should we have a cancel?

app/cho/shared/deletes_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  # DELETE /deletes
+  # @note The flash hash is used to pass some data to the index view and display a list of the
+  #       deleted titles. This could be expanded later to report any resources that were not
+  #       successfully deleted.
+  def destroy
+    resources.each do |resource|
+      change_set_persister.delete(change_set: DeleteSet.new(resource))
+    end
+    flash[:deleted_titles] = resources.map(&:title).flatten
```
cam156: Would it be better to send the titles as a parameter to the index? and or just render a response without redirecting?

config/routes.rb
```diff
@@ -25,12 +25,16 @@
...
+  resources :library_collections, except: [:show, :index, :destroy], controller: 'collection/library_collections'
+  resources :curated_collections, except: [:show, :index, :destroy], controller: 'collection/curated_collections'
   resources :data_dictionary_fields, controller: 'data_dictionary/fields'
   resources :work_import_csvfile, as: 'csv_file', path: '/csv_file', only: [:new, :create], controller: 'work/import/csv'
 
+  get '/deletes', to: 'deletes#index'
+  post '/deletes', to: 'deletes#confirm'
+  delete '/deletes', to: 'deletes#destroy'
+
   post '/csv_file/run_import', to: 'work/import/csv#run_import'
```
cam156: This line should be kept with the other csv items above

spec/cho/shared/deletes_controller_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    it 'destroys the requested resources' do
+      expect {
+        delete :destroy, params: { delete: { ids: [work.to_param, collection.to_param] } }
+      }.to change { metadata_adapter.query_service.find_all.to_a.count }.by(-2)
+      expect(flash[:deleted_titles]).to eq([work.title.first, collection.title.first])
+    end
+  end
+
+  describe 'GET #index' do
```
cam156: This really does not test anything.  I think it is becuase forwarding to the index is not really what you want.

app/cho/shared/deletes_controller.rb
awead: @cam156 that would pass the parameters via the URL. If we end up deleting a lot of titles, that would make the url too long. I'm trying to follow a REST/Rails idiom for deletes, where the deletes action is performed and then there's a redirect. This seems to follow other actions where something created or updated, then directed to another page. I supposed we could delete the items and render the page in the same action, but that seems wrong.

app/views/deletes/confirm.html.erb
awead: @cam156 I wondered that too, but thought it would just be a back button, and we don't want those, if I recall. Where else would cancel go?

app/valkyrie/change_set_persister.rb
awead: @cam156 I wonder also if the whole process of deleting the binary should be on the resources as well, ex. `resource.delete_file` or something.

awead: @cam156 but that seems to break the functional pattern.

app/cho/shared/deletes_controller.rb
awead: @cam156 the index action is rendering the final results of the delete. There's a view for it, so there needs to be a defined action, but Rail might render the view even without the action defined, if there's a route for it.

spec/cho/shared/deletes_controller_spec.rb
awead: @cam156 see my comment above. I think this sticks to the Rails pattern of create/update/delete then redirect.

app/cho/shared/deletes_controller.rb
awead: @cam156 how is that any different then passing `resources.count`, and saying "You've successfully deleted 3 items" ? We could change the message to that, if you prefer.

spec/cho/shared/deletes_controller_spec.rb
awead: @cam156 If there's no view file present, it fails.

app/cho/shared/deletes_controller.rb
cam156: @awead I guess I question the need for the fake index action.  What about a flash message and then redirect to the home page?  Then we do not need to maintain the index action that does nothing?

app/views/deletes/confirm.html.erb
cam156: I think rails give you access to the page referer `request.referer` and just use that for the url for the cancel button, instead of using javascript.

spec/cho/shared/deletes_controller_spec.rb
cam156: @awead ok, but the view could be completely empty and the test would still pass.

awead:
@cam156 I've updated this with the additional changes from #450 since it was tightly related to this PR.

Issue #475, opened 2018-04-12
Enforce the text type in data dictionary for description


awead:
The data dictionary defines the field type for description as "text" instead of "string" for the others. This implies a distinction between an `input` html type for strings, and `textarea` for the "text" type.

This needs to occur for both works and collections.

PR #504, opened 2018-05-02
Chose this PR to enforce the text type in data dictionary for description

awead:
# Option A

## Description

Renders a text area when the data dictionary field type is 'text.' By
default, everything else is still a text input.

Fixes: #475 

Why was this necessary? The description field was not rendered as a text area.

## Changes

## Checklist

- [ ] i18n enabled
- [ ] adequate test coverage
- [ ] accessible interface

### DIFF
--- a/app/cho/schema/metadata_field.rb
+++ b/app/cho/schema/metadata_field.rb
@@ -10,5 +10,13 @@ module Schema
       field.data_dictionary_field_id = data_dictionary_field.id
       field
     end
+
+    def input_type
+      if text?
+        :text_area
+      else
+        :text_field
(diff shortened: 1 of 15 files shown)


app/cho/schema/metadata_field.rb
```diff
@@ -10,5 +10,13 @@ def self.initialize_from_data_dictionary_field(data_dictionary_field)
       field.data_dictionary_field_id = data_dictionary_field.id
       field
     end
+
+    def input_type
+      if field_type == 'text'
```
awead: I think this can just be `text?`

PR #515, opened 2018-05-04
Allow repository resets in selected service environments

awead:
## Description

Cleaning out the repository and running benchmarks is a common task for specific environments. However, it should only be allowed in specific ones, for example, only in QA or local and not in production.

Why was this necessary?

This makes running benchmarks and resetting the QA environment an easier process.

## Changes

Are there any major changes in this PR that will change the release process? Yes, we need to add a new key to the server's application.yml file.

## Checklist

- [ ] i18n enabled
- [ ] adequate test coverage
- [ ] accessible interface

### DIFF
--- a/app/cho/metrics/repository.rb
+++ b/app/cho/metrics/repository.rb
@@ -1,0 +1,19 @@
+# frozen_string_literal: true
+
+module Metrics
+  class Repository
+    WHITELIST = ['localhost', 'qa'].freeze
+
+    class ResetError < StandardError; end
+
+    def self.reset
+      raise ResetError, "Reset not allowed for #{ENV['service_name']}" unless WHITELIST.include?(ENV['service_name'])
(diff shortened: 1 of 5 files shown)


app/cho/metrics/repository.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Metrics
```
cam156: @awead this seems more general purpose than just metric.  I'm wondering on the namespace.

awead: @cam156 it's meant to be used specifically with metrics, such as before running a benchmark, but it's not exclusive to metrics. I put it there because that's what I was using it for. I'm open to other suggestions. 😄

cam156: ?shared?  Yes it is not shared yet, but it would be sharable.  We can leave it in metrics.  It just seemed more general...

Issue #466, opened 2018-04-11
Upgrade to Valkyrie 1.0.0


awead:
This version no longer supports persisting non-array attributes, which could present a problem.

PR #520, opened 2018-05-07
Creating persisters that allow for singular, since it works for us

cam156:
## Description

I was walking down the path of making everything plural and then I got to the part where we were outputting our data and realized it really needed to be singular there.  That meant we would need to translate from plural to singular.  That got me thinking why translate at the outside of the system.  Why not do the translation at the persistence layer?

I brought that thought to refactoring and we took it a step further.  What would happen if we just allowed Singular to  work.  The result is a set of persister that do not check for multiple.

This made almost every test pass, except for the few that relied on the new parameter new_record.

I'm not entirely positive this is what we want to do, or if we want to create an "Saveable" object that translates any singular property into multiple before it is saved.

References ticket: #466

Why was this necessary?  Upgrade to Valkyrie 1.0 to upgrade to blacklight 7

## Changes

* Upgrade to Valkyrie 1.0
* Add Singular Adapters & Persisters to allow for singular values
* Add new_record where needed in the tests

## Checklist

- [ ] i18n enabled
- [x] adequate test coverage
- [ ] accessible interface

### DIFF
--- a/app/valkyrie/memory/singular_persister.rb
+++ b/app/valkyrie/memory/singular_persister.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+
+class Memory::SingularPersister < Valkyrie::Persistence::Memory::Persister
+  # overriding ensure_multiple_values! to do nothing since we are allowing singular values
+  def ensure_multiple_values!(resource); end
+end
--- a/app/valkyrie/postgres/singular_persister.rb
+++ b/app/valkyrie/postgres/singular_persister.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+module Postgres
+  class SingularPersister < Valkyrie::Persistence::Postgres::Persister
+    private
+
+      # overriding ensure_multiple_values! to do nothing since we are allowing singular values
+      def ensure_multiple_values!(resource); end
+  end
+end
--- a/app/valkyrie/solr/singular_repository.rb
+++ b/app/valkyrie/solr/singular_repository.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+
+class Solr::SingularRepository < Valkyrie::Persistence::Solr::Repository
+  # overriding ensure_multiple_values! to do nothing since we are allowing singular values
+  def ensure_multiple_values!(resource); end
+end
--- a/spec/valkyrie/indexing_adapter_spec.rb
+++ b/spec/valkyrie/indexing_adapter_spec.rb
@@ -5,14 +5,23 @@ require 'valkyrie/specs/shared_specs'
 
 RSpec.describe IndexingAdapter do
   let(:adapter) do
-    described_class.new(metadata_adapter: Valkyrie::Persistence::Memory::MetadataAdapter.new,
+    described_class.new(metadata_adapter: Memory::SingularMetadataAdapter.new,
                         index_adapter: index_solr)
   end
   let(:query_service) { adapter.query_service }
   let(:persister)     { adapter.persister }
   let(:index_solr)    { Valkyrie::MetadataAdapter.find(:index_solr) }
 
-  it_behaves_like 'a Valkyrie::Persister'
+  context 'shared examples allowing singular values' do
+    before do |example|
+      # We are skipping the check for 'does not save non-array properties' since we are allowing singular values
+      if example.description == 'does not save non-array properties'
+        skip 'We are allowing Non Array Properties'
+      end
+    end
+
+    it_behaves_like 'a Valkyrie::Persister'
+  end
 
   it 'updates Solr after saving to the metadata adapter' do
     persister.buffer_into_index do |buffered_adapter|
--- a/spec/valkyrie/memory/singular_persister_spec.rb
+++ b/spec/valkyrie/memory/singular_persister_spec.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require 'valkyrie/specs/shared_specs'
+
+RSpec.describe Memory::SingularPersister do
+  let(:persister) { described_class.new(adapter) }
+  let(:adapter) { Memory::SingularMetadataAdapter.new }
+  let(:query_service) { adapter.query_service }
+
+  context 'shared examples allowing singular values' do
+    before do |example|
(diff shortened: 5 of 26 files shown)


app/valkyrie/memory/singular_persister.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class Memory::SingularPersister < Valkyrie::Persistence::Memory::Persister
+  def ensure_multiple_values!(resource)
+    # todo We should change singular to multiple or something
```
awead: @cam156 is this something we need to do later, or still have to decide to?

app/valkyrie/postgres/singular_persister.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Postgres
+  class SingularPersister < Valkyrie::Persistence::Postgres::Persister
+    private
+
+      def ensure_multiple_values!(resource)
+        # todo We should change singular to multiple or something
```
awead: Same comment here. Maybe say something like, we're overriding all these persisters so we can have singular attributes, but we may want to change this later actually persist multiple values instead. Here's how this could work...

app/valkyrie/solr/singular_repository.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class Solr::SingularRepository < Valkyrie::Persistence::Solr::Repository
+  def ensure_multiple_values!(resource)
+    # todo We should change singular to multiple or something
```
awead: Same comment here. We should remind ourselves that we're doing this to override Valkyrie's now default behavior of persisting only arrays.

spec/valkyrie/indexing_adapter_spec.rb
```diff
@@ -5,14 +5,23 @@
...
+    described_class.new(metadata_adapter: Memory::SingularMetadataAdapter.new,
                         index_adapter: index_solr)
   end
   let(:query_service) { adapter.query_service }
   let(:persister)     { adapter.persister }
   let(:index_solr)    { Valkyrie::MetadataAdapter.find(:index_solr) }
 
-  it_behaves_like 'a Valkyrie::Persister'
+  context 'shared examples' do
+    # We are not going to run the check for saving arrays since we are allowing it
```
awead: It looks like this is overriding the share spec, right? Could you update the comment to reflect that?

spec/valkyrie/memory/singular_persister_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+require 'rails_helper'
+require 'valkyrie/specs/shared_specs'
+
+RSpec.describe Memory::SingularPersister do
+  let(:persister) { described_class.new(adapter) }
+  let(:adapter) { Memory::SingularMetadataAdapter.new }
+  let(:query_service) { adapter.query_service }
+
+  context 'shared examples' do
+    # We are not going to run the check for saving arrays since we are allowing it
```
awead: Ditto here. Maybe even a shared local spec that overrides the Valkyrie shared spec? Don't know if that's possible...

spec/valkyrie/postgres/singular_persister_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+require 'rails_helper'
+require 'valkyrie/specs/shared_specs'
+
+RSpec.describe Postgres::SingularPersister do
+  let(:persister) { described_class.new(adapter: adapter) }
+  let(:adapter) { Postgres::SingularMetadataAdapter.new }
+  let(:query_service) { adapter.query_service }
+
+  context 'shared examples' do
+    # We are not going to run the check for saving arrays since we are allowing it
```
awead: Ditto here.

spec/valkyrie/solr/singular_persister_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+RSpec.describe Solr::SingularPersister do
+  subject { persister }
+
+  let(:persister) { described_class.new(adapter: adapter) }
+  let(:adapter) { Solr::SingularMetadataAdapter.new(connection: Blacklight.default_index.connection) }
+  let(:query_service) { adapter.query_service }
+
+  context 'shared examples' do
+    # We are not going to run the check for saving arrays since we are allowing it
```
awead: Here too, but looks to be slightly different? Maybe the shared override doesn't work here.

awead:
@cam156 I had just some general comments on the comments themselves, for the benefit of our future selves, but overall, this looks great to me!

spec/valkyrie/indexing_adapter_spec.rb
cam156: @awead we are just skipping a spec, which is what the comment says.  Not sure what you are getting at.

app/valkyrie/memory/singular_persister.rb
cam156: @awead Not sure.  It's really a talking point. We could be happy with the solution (becuase it works) and I could change the comment to `overriding ensure_multiple_values, since we do not require multiple to save` or something like that, or we could leave in the TODO if that is a direction we want to pursue, and create a ticket for it.

cam156:
@awead updated some of the comments. Added an additional test so that the PR did not decrease coverage.  Maybe I hit what you were asking for in the comments, but I'm not certain I got what you were asking.

spec/valkyrie/indexing_adapter_spec.rb
awead: @cam156 we're skipping the spec that comes from "a Valkyrie::Persister" right? That's what I meant by "override" because that bit of code comes from the Valkyrie gem. It's not really overriding, but it was unclear to me at first that's what was happening. Not a big deal, though.

app/valkyrie/solr/singular_repository.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+class Solr::SingularRepository < Valkyrie::Persistence::Solr::Repository
+  # overriding ensure_multiple_values! to do nothing since we are allowing singular values
```
awead: @cam156 yep, much better!

awead:
@cam156 yeah, I think this is fine now.

PR #521, opened 2018-05-07
Submission change sets should require a collection

awead:
## Description

A submission change change with an empty array of member collections will still be valid. This seems wrong. If works are supposed to always be in a collection, then the submission change set should not be valid if there is not at least one id in the array of member collections.

Why was this necessary?

This should prevent the occurrence of works without a parent collection.

## Changes

## Checklist

- [ ] i18n enabled
- [ ] adequate test coverage
- [ ] accessible interface

### DIFF
--- a/spec/cho/work/submissions/change_set_spec.rb
+++ b/spec/cho/work/submissions/change_set_spec.rb
@@ -69,16 +69,17 @@ RSpec.describe Work::SubmissionChangeSet do
 
       its(:full_messages) { is_expected.to include('Work type bad one does not exist') }
     end
-    context 'with all required fields' do
-      let(:params) { { work_type_id: work_type_id, title: 'Title' } }
-
-      its(:full_messages) { is_expected.to be_empty }
-    end
 
     context 'with non-existent parents' do
       let(:params) { { work_type_id: work_type_id, title: 'Title', member_of_collection_ids: ['nothere'] } }
 
       its(:full_messages) { is_expected.to include('Member of collection ids nothere does not exist') }
+    end
+
+    context 'without collection membership' do
+      let(:params) { { work_type_id: work_type_id, title: 'Title', member_of_collection_ids: [] } }
+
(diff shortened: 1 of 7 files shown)


spec/cho/work/submissions/change_set_spec.rb
```diff
@@ -69,18 +69,19 @@
 
       its(:full_messages) { is_expected.to include('Work type bad one does not exist') }
     end
-    context 'with all required fields' do
```
cam156: I would expect this check to remain and include a collection so we can see one that validates.

spec/valkyrie/change_set_persister_spec.rb
```diff
@@ -42,13 +42,14 @@
...
-      let(:resource)   { build(:work, title: 'with a file') }
+      let!(:collection) { create(:collection) }
+      let(:resource) { build(:work, title: 'with a file', member_of_collection_ids: [collection.id]) }
       let(:change_set) { Work::SubmissionChangeSet.new(resource) }
       let(:temp_file) { Rack::Test::UploadedFile.new(Rails.root.join('spec', 'fixtures', 'hello_world.txt')) }
       let(:work_files) { Work::File.all }
       let(:work_file) { work_files.first }
 
-      it 'saves the record and the file' do
+      it 'saves the record, its collection and file' do
```
cam156: @awead you are not saving the collection.  Is this really accurate?

spec/cho/work/submissions/change_set_spec.rb
awead: @cam156 there's a check for all fields down on line 85. I think this was an old test prior to when we had collections.

spec/valkyrie/change_set_persister_spec.rb
awead: @cam156 ah no, I was working on the test and changing the wording, but forgot to re-change the wording once I had it figured out. The test will not create a collection, just the file and work. I'll push up a change.

Issue #437, opened 2018-03-09
Provide new values for metadata fields


awead:
Fields that are copied from the data dictionary into metadata fields can have their values overridden. This currently could take place in config/data_dictionary/schema_fields.yml

awead:
We'd like to add a new display name to a schema field in `schema_fields.yml`. A good example would be making the contributor field in the document schema have the display name of "Author" instead of just "Contributor." It should also work for all the other options in the data dictionary, such as required designation, multiple, etc.

The code changes would take place in seeds.rb.

awead:
@cam156 I tried demoing this at the retrospective, but wasn't able. I tried updating the schema_fields.yml file and re-seeding the application.

PR #528, opened 2018-05-10
Moving code out of seed into appropriate classes

cam156:
This allows it to be tested

This is a precursor to adding more functionality to the Schema configuration to allow for each schema to override data dictionary field attribute values

refs #437

## Description

References ticket: (if any)

Why was this necessary?

## Changes

## Checklist

- [ ] i18n enabled
- [ ] adequate test coverage
- [ ] accessible interface

### DIFF
--- a/app/cho/data_dictionary/csv_importer.rb
+++ b/app/cho/data_dictionary/csv_importer.rb
@@ -6,6 +6,12 @@ module DataDictionary
   class CsvImporter
     attr_reader :file, :lines, :attributes
 
+    def self.load_dictionary(dictionary_file = nil)
+      dictionary_file ||= Rails.root.join('config', 'data_dictionary', "data_dictionary_#{Rails.env.downcase}.csv")
+      importer = new(File.new(dictionary_file))
+      importer.import
+    end
+
     def initialize(file)
       @file = file
     end
--- a/app/cho/schema/configuration.rb
+++ b/app/cho/schema/configuration.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+module Schema
+  class Configuration
+    attr_reader :schema_config, :work_types
+
+    def initialize(schema_file_path = nil)
+      schema_file_path ||= Rails.root.join('config', 'data_dictionary', 'schema_fields.yml')
+      @schema_config = YAML.load_file(schema_file_path)[Rails.env]
+      @work_types = schema_config.map { |config| config.fetch('schema') }
+    end
+
+    def core_field_count
+      core_field_ids.length
+    end
+
+    def core_field_ids
+      @core_field_ids ||= Schema::MetadataCoreFields.generate(Valkyrie.config.metadata_adapter).map(&:id)
+    end
+
+    def load_work_types
+      work_types.each do |type|
+        work_type_config = Schema::WorkTypeConfiguration.new(schema_configuration: self, work_type: type)
+        work_type_config.generate_metadata_schema
+        work_type_config.generate_work_type unless type == 'Collection'
+      end
+    end
+  end
+end
--- a/app/cho/schema/work_type_configuration.rb
+++ b/app/cho/schema/work_type_configuration.rb
@@ -1,0 +1,65 @@
+# frozen_string_literal: true
+
+module Schema
+  class WorkTypeConfiguration
+    attr_reader :schema_configuration, :work_type, :fields
+
+    def initialize(schema_configuration:, work_type:)
+      @schema_configuration = schema_configuration
+      @work_type = work_type
+      @fields = initialize_fields
+    end
+
+    def generate_metadata_schema
+      schema_fields = initialize_schema_fields
+
+      metadata_schema = Schema::Metadata.new(
+        label: work_type,
+        core_fields: schema_configuration.core_field_ids,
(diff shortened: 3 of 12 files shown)


app/cho/data_dictionary/csv_importer.rb
```diff
@@ -6,6 +6,12 @@ module DataDictionary
   class CsvImporter
     attr_reader :file, :lines, :attributes
 
+    def self.load_dictionary(dictionary_file = nil)
+      dictionary_file ||= Rails.root.join('config', 'data_dictionary', "data_dictionary_#{Rails.env.downcase}.csv")
+      importer = DataDictionary::CsvImporter.new(File.new(dictionary_file))
```
awead: This can be just `new` or `self.new` here.

app/cho/schema/configuration.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module Schema
+  class Configuration
+    attr_reader :schema_config, :work_types
+
+    def initialize(schema_file_path = nil)
+      schema_file_path ||= Rails.root.join('config', 'data_dictionary', 'schema_fields.yml')
+      @schema_config ||= YAML.load_file(schema_file_path)[Rails.env]
+      @work_types ||= schema_config.map { |config| config.fetch('schema') }
```
awead: Why do you need to `||=` these two instance variables?

app/cho/schema/work_type_configuration.rb
```diff
@@ -0,0 +1,66 @@
...
+      work_type_schema = Work::Type.new(
+        label: work_type,
+        metadata_schema_id: metadata_schema.id,
+        processing_schema: 'default',
+        display_schema: 'default'
+      )
+      find_or_save(work_type_schema)
+    end
+
+    def initialize_schema_fields
```
awead: Should this be a private method? I don't see it used outside the class.

app/cho/schema/work_type_configuration.rb
```diff
@@ -0,0 +1,66 @@
...
+      fields_for_type = fields
```
awead: Why re-assign `fields` to `fields_for_type`? You could stick with the `fields` instance variable or perhaps rename it to `fields_for_type` in the initializer if that has a better fit.

app/valkyrie/change_set_persister.rb
```diff
@@ -14,6 +14,12 @@ def save(change_set:)
     persister.save(resource: change_set.resource)
   end
 
+  def find_or_save(resource, unique_attribute:)
```
awead: This reads more like `find_or_update`.

app/valkyrie/change_set_persister.rb
```diff
@@ -14,6 +14,12 @@ def save(change_set:)
...
+    attribute_value = resource.send(unique_attribute)
+    return resource.class.where(unique_attribute => attribute_value).first if resource.class.where(label: attribute_value).count.positive?
```
awead: I wonder if we should make an `exists?` class method on our resources to make this a bit easier, on some other AR pattern.

spec/cho/schema/configuration_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  its(:schema_config) { is_expected.to eq([{ 'schema' => 'Generic', 'fields' => { 'generic_field' => { 'order_index' => 1 } } },
+                                           { 'schema' => 'Document', 'fields' => { 'document_field' => { 'order_index' => 1 } } },
+                                           { 'schema' => 'Still Image',
+                                             'fields' => { 'still_image_field' => { 'order_index' => 1 } } },
+                                           { 'schema' => 'Map', 'fields' => { 'map_field' => { 'order_index' => 1 } } },
+                                           { 'schema' => 'Moving Image',
+                                             'fields' => { 'moving_image_field' => { 'order_index' => 1 } } },
+                                           { 'schema' => 'Audio', 'fields' => { 'audio_field' => { 'order_index' => 1 } } },
+                                           { 'schema' => 'Collection', 'fields' => [] }])}
```
awead: If you prefer, another formatting option is:
``` ruby
its(:schema_config) { is_expected.to eq(
  [
    { 'schema' etc... }
```
Rubocop won't complain about that one.

app/valkyrie/change_set_persister.rb
cam156: @awead the most accurate would be `find_and_update_or_create`, but that is a mouthful. It either Finds and updates, or creates.

app/cho/schema/work_type_configuration.rb
cam156: @awead I left it as a public method so I could test it.  It could be a private method, but that made testing a bit harder.

spec/cho/schema/configuration_spec.rb
cam156: @awead rubocop is not upset about the current one.  I can change it if you want me to, but I kinda like the expect indented.

awead: @cam156 I know it's allowing the current one. The problem is, content is pushed off so far to the right, it's not readable unless you scroll over. We can keep it for now, but I'm considering updating our style guidelines to not allow for long lines.

PR #557, opened 2018-06-19
FileSet resource class for grouping files

### LINKED ISSUE #550: Create FileSet resource
Valkyrie resource for a file set. 
  * has a property for containing many Work::File resources
  * has its own schema for defining its own properties, ex. display name

Child of #545

awead:
## Description

Fixes #550 

Provides a resource to group multiple Work::File objects together. This is analogous to the FileSet class in the Hydra works extension to PCDM. One original file is grouped together with multiple derivative files, including auto and manually generated derivatives, and extracted information such as text or other technical metadata.

## Changes

## Checklist

- [x] i18n enabled
- [x] adequate test coverage
- [x] accessible interface

### DIFF
--- a/app/cho/schema/configuration.rb
+++ b/app/cho/schema/configuration.rb
@@ -2,12 +2,16 @@
 
 module Schema
   class Configuration
-    attr_reader :schema_config, :work_types
+    attr_reader :schemas, :schema_config
+
+    Struct.new('Schema', :type, :work_type?)
 
     def initialize(schema_file_path = nil)
       schema_file_path ||= Rails.root.join('config', 'data_dictionary', 'schema_fields.yml')
       @schema_config = YAML.load_file(schema_file_path)[Rails.env]
-      @work_types = schema_config.map { |config| config.fetch('schema') }
+      @schemas = schema_config.map do |config|
+        Struct::Schema.new(config.fetch('schema'), (config.fetch('work_type', 'true') == 'true'))
+      end
     end
 
     def core_field_count
@@ -19,10 +23,10 @@ module Schema
     end
 
     def load_work_types
-      work_types.each do |type|
-        work_type_config = Schema::WorkTypeConfiguration.new(schema_configuration: self, work_type: type)
+      schemas.each do |schema|
+        work_type_config = Schema::WorkTypeConfiguration.new(schema_configuration: self, work_type: schema.type)
         work_type_config.generate_metadata_schema
-        work_type_config.generate_work_type unless type == 'Collection'
+        work_type_config.generate_work_type if schema.work_type?
       end
     end
   end
--- a/app/cho/work/file_set_change_set.rb
+++ b/app/cho/work/file_set_change_set.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+module Work
+  class FileSetChangeSet < Valkyrie::ChangeSet
+    validates :member_ids, with: :validate_members!
+    validates :member_ids, presence: true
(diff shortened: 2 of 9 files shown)


spec/cho/schema/configuration_spec.rb
```diff
@@ -19,15 +19,12 @@
...
           'fields' => { 'moving_image_field' => { 'order_index' => 1 } } },
         { 'schema' => 'Audio', 'fields' => { 'audio_field' => { 'order_index' => 1 } } },
-        { 'schema' => 'Collection', 'fields' => [] }
+        { 'schema' => 'Collection', 'fields' => [], 'work_type' => 'false' },
+        { 'schema' => 'FileSet', 'fields' => [], 'work_type' => 'false' }
       ]
     )
   end
 
-  its(:work_types) do
```
cam156: This test should be replaced with a test for schemas that you are expecting.

app/cho/schema/configuration.rb
```diff
@@ -2,12 +2,16 @@
...
+    attr_reader :schemas, :schema_config
+
+    Struct.new('Schema', :type, :fields, :work_type?)
 
     def initialize(schema_file_path = nil)
       schema_file_path ||= Rails.root.join('config', 'data_dictionary', 'schema_fields.yml')
       @schema_config = YAML.load_file(schema_file_path)[Rails.env]
-      @work_types = schema_config.map { |config| config.fetch('schema') }
+      @schemas = schema_config.map do |config|
+        Struct::Schema.new(config.fetch('schema'), config.fetch('fields'), config.fetch('work_type', true))
```
cam156: Why are you saving off fields?  Nothing is being done with them in this code.

app/cho/work/file_set_change_set.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Work
```
cam156: Should this live in the work module or have it's own module?

awead: @cam156 We're already putting File in the Work module, so FileSet and its change set made sense. If there were to be a new module, I could see File, FileSet, FileChangeSet, etc. all going in there, but what would we call it?

cam156: Good Question.   We can leave it here for now and move it if we end up using the FileSet for other things besides the work.  I knew the File stuff was just a place holder when I put it in so I would not lean on that as having any thought behind it.  I just stuffed it some place until we knew what we were doing with files.

awead: @cam156 yes, and the file code will grow, so it should be moved out at some point, or reorganized. I just don't have a clear idea yet as to what the module/concept would be.

spec/cho/schema/configuration_spec.rb
```diff
@@ -19,13 +19,19 @@
...
-    is_expected.to eq(['Generic', 'Document', 'Still Image', 'Map', 'Moving Image', 'Audio', 'Collection'])
+  describe '#schemas' do
+    it 'reads each schema from the config file' do
+      expect(schema_configuration.schemas.first).to be_a(Struct::Schema)
+      expect(schema_configuration.schemas.map(&:type)).to contain_exactly(
+        'Generic', 'Document', 'Still Image', 'Map', 'Moving Image', 'Audio', 'Collection', 'FileSet'
+      )
```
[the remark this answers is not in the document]
awead: @cam156 good catch. There was an issue with `work_type?` not returning a boolean.

Issue #512, opened 2018-05-04
Define a schema for files


awead:
This would define properties such as its type. Ex, preservation, service, thumbnail, etc. We should use PCDM file type extensions for this.
 
Questions:
* having metadata fields in a schema that are not in the data dictionary?
* adding technical and rights metadata? (look to Premis for predicates)

Child of #545

awead:
Define use type field in data dictionary and allow core fields on files in the schema definition. The core fields won't be used on the file resources in the application.

PR #567, opened 2018-06-21
Add use type URIs to Work::File

awead:
## Description

Connects to #512

This doesn't define a schema for Work::File resources, but instead provides an internal controlled vocabulary for what use types we can assign to the file.

I'm totally satisfied with it. I've provided a Work::FileChangeSet to provided better feedback for errors with file, but we're not using it yet. I imagine it will be needed for later tickets when reporting errors with file sets during the bag import process.

Also, there is duplication of validation between the file resource and its change set. This is so the existing change set persister can save files correctly, with a default preservation use type. This will also need to change later as we will want to pass specific use types to the persister.

## Changes

## Checklist

- [ ] i18n enabled
- [ ] adequate test coverage
- [ ] accessible interface

### DIFF
--- a/app/cho/work/use_type_validator.rb
+++ b/app/cho/work/use_type_validator.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+module Work
+  class UseTypeValidator < ActiveModel::EachValidator
+    def validate_each(record, attribute, value)
+      values = Array.wrap(value)
+      bad_uris = values.reject { |v| Work::WithUseType::USE_TYPE_URIS.include?(v) }
+      return if bad_uris.empty? && values.present?
+      record.errors.add attribute, error_message(bad_uris)
+    end
+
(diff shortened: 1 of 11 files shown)


app/cho/work/use_type_validator.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Work
+  class UseTypeValidator < ActiveModel::EachValidator
+    def validate_each(record, attribute, value)
```
cam156: Would it be possible to just validate this against the list? something like
` validates :index_type, inclusion: { in: DataDictionary::Field::IndexTypes.values }`

awead: @cam156 That was one of my earlier attempts. It didn't seem to work because the attribute itself is an array and it can't verify that an array as elements that are included in another array. I wish it could!

Issue #551, opened 2018-06-18
Create a processor class


awead:
Controls how files are processed. They are attached to a work type and include steps for actions such as characterization with FITS and text extraction with Tika.

Details:
* investigate existing patterns, including dry-ruby, and Tom Johnson's processor chain work in Hydra
* modify Work::Type to include an array of processors
* define the base class that will eventually call out to the different processors like 
* defines a "runner" that executes the task

Child of #545

cam156:
Looked at both the new code in Hyrax and dry-rb and both are utilizing http://dry-rb.org/gems/dry-transaction/ for a similar desire.

awead:
Implementation details:
* implement dry-transactions to ingest new files into CHO
* updating, replacing, or deleting files can be addressed later (we can leave in the operation classes, though)
* creation operation should take one _or more_ files, put them into a FileSet and add them to the work
  * Sub-task: update the Work resource have file set member ids
  * Sub-task: move files attributes and other Shrine-related code from Work::Submission to Work::FileSet
  * Multi-file upload via GUI is out of scope
* add the file set to the work
* remove file-creation code from the ChangeSetPersister

cam156:
This ticket was a team effort. @awead finished up the processing class PR, which was merged this morning.

PR #576, opened 2018-06-27
Adding in dry-transactions for creating updating and deleting a file

cam156:
## Description

@awead  @mtribone and I have been looking at how dry-transactions could fit into the system.

We included the gem and added classes (no implementation), and configuration for how it might work.  Let us know if this seems reasonable and we will try to hook it into the file creation process in the existing system.

The very basic idea is that you have Transactions in the system that you can choose from in your configuration for what you want to occur when a file is created, updated, and deleted.  Each transaction is a dry-transaction, which is a series of operations.  Operations cab be reused between Transactions and we could have different Transactions for different work types.

## Changes

Adds a basic set of dry-transactions operations for a file.

## Checklist

- [ ] i18n enabled
- [ ] adequate test coverage
- [ ] accessible interface

### DIFF
--- a/app/valkyrie/change_set_persister.rb
+++ b/app/valkyrie/change_set_persister.rb
@@ -78,27 +78,11 @@ class ChangeSetPersister
 
   private
 
+    # @return [Valkyrie::ChangeSet] change set for the resource, such as a work or collection
     def store_files(change_set)
-      return change_set unless change_set.respond_to?(:file) && change_set.file.present?
-      adapter_file = adapter_file(change_set: change_set)
-      saved_work_file = persister.save(resource: work_file(change_set: change_set, adapter_file: adapter_file))
-      change_set.model.files << saved_work_file.id
-      change_set
-    end
-
-    def work_file(change_set:, adapter_file:)
-      Work::File.new(
-        file_identifier: adapter_file.id,
-        original_filename: change_set.file.original_filename
-      )
-    end
-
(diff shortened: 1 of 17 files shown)


awead:
@cam156 I'm 👍 to the overall idea. Leaning on the dry framework should be preferable to rolling our own. We can talk specifics at standup tomorrow.

awead:
@cam156 I'm pretty sold on this approach. The docs recommend using dry-system, "to make it easier to populate a container with your own objects." Is this voodoo magic or something to consider?

app/valkyrie/change_set_persister.rb
```diff
@@ -78,27 +78,11 @@ def validate_and_save_with_buffer(change_set:, resource_params:)
...
-    end
-
-    def adapter_file(change_set:)
-      storage_adapter.upload(
-        file: change_set.file.tempfile,
-        original_filename: change_set.file.original_filename,
-        resource: change_set.model
-      )
+      result = Transaction::File::Create.new.call(change_set)
+      return result.failure if result.failure?
```
cam156: @awead prior to this change the result was always the change set.  Now it is returning something else if there is a failure.  Will saving a collection still work?

awead: @cam156 yes, `result.failure` is still the change set. However, maybe `return change_set if result.failure?` would be clearer.

awead:
Fixes #551

PR #577, opened 2018-06-27
Validating bags for import

### LINKED ISSUE #561: Validate work bags
Validate our bags so that they conform to the specifications in https://github.com/psu-libraries/cho/wiki/File-Specifications as well checksum validation.

This includes abilities to query bags for number of file sets, preservations files, service files, and the tallying of errors associated with that.

Child of #545

awead:
## Description

Fixes: #561 

Validates the structure and integrity of bags containing one or more works, each will multiple files and file sets.

While it will parse bags of nested works, it doesn't yet validate these if the nested works have errors. This test has been marked pending and can be addressed when nested works are added later.

Tests should cover all the cases outlined in the [file specs](https://github.com/psu-libraries/cho/wiki/File-Specifications)

Connected to #561


app/cho/bag/file.rb
```diff
@@ -0,0 +1,72 @@
...
+    'access' => Vocab::FileUse.AccessFile,
+    'thumb' => Valkyrie::Vocab::PCDMUse.ThumbnailImage,
+    'front' => Vocab::FileUse.FrontImage,
+    'back' => Vocab::FileUse.BackImage
+  }.freeze
+
+  # @param [Pathname] file
+  def initialize(file)
+    @file = file
+    @parts = file.basename(File.extname(file)).to_s.split(/_/)
```
cam156: Using just  '_' is slightly more efficient (and in my mind easier to read)
```
[47] pry(main)> now = DateTime.now; (1..1000).each  {|i|path.basename(File.extname(path)).to_s.split(/_/)}; puts DateTime.now.to_f - now.to_f
0.011595964431762695
=> nil
[48] pry(main)> now = DateTime.now; (1..1000).each  {|i|path.basename(File.extname(path)).to_s.split('_')}; puts DateTime.now.to_f - now.to_f
0.007995843887329102
```

app/cho/bag/file_validator.rb
```diff
@@ -0,0 +1,24 @@
...
+  class FileValidator < ActiveModel::Validator
+    def validate(record)
+      validate_object_id(record)
+      validate_type(record)
+    end
+
+    private
+
+      def validate_object_id(record)
+        return if record.object_id == record.file.parent.basename.to_s
```
cam156: Is this assuming a single file per record?  Is a record a work or a file set?

app/cho/bag/work.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  def file_sets
+    file_set_ids = files.map(&:file_set_id).uniq
+    file_set_ids.map do |id|
+      FileSet.new(files.select { |file| file.file_set_id == id })
+    end
+  end
+
+  class FileSet
```
cam156: I think this should be in it's own file not relative to the work.

app/cho/bag/work_validator.rb
```diff
@@ -0,0 +1,29 @@
...
+  class WorkValidator < ActiveModel::Validator
+    def validate(record)
+      record.files.reject(&:valid?).each do |file|
+        file.errors.full_messages.each do |error|
+          record.errors.add(:files, error)
+        end
+      end
+
+      record.file_sets.each do |file_set|
+        if file_set.files.select(&:service?).empty? && file_set.files.select(&:preservation?).empty?
```
cam156: How dose a representative relate to service and preservation?  Can yo have all 3? If there is a service and a preservation, but no representative should an error be produced?

spec/cho/bag/package_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+
+require 'rails_helper'
+
+RSpec.describe Bag::Package do
+  describe 'a valid bag' do
+    subject(:package) do
+      BagFactory::Package.create(
+        batch_id: 'batchID_2008-06-23',
+        data: {
+          objectID: [
```
cam156: These look like file names, not objectIDs

spec/cho/bag/work_spec.rb
```diff
@@ -0,0 +1,313 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Bag::Work do
+  describe 'a simple work' do
+    subject(:work) do
+      BagFactory::Work.new_create(
+        objectID: [
```
cam156: Are these really objects?

carolyncole (changes requested):
@awead overall it looks good.  I'm slightly confused about the concept of Package.  Is that just the top level of the bag?

app/cho/bag/file.rb
awead: @cam156 huh. I never knew you could pass a string to `split`. I always thought it had to be a regex.

app/cho/bag/file_validator.rb
awead: `record` here is a file, or `Bag::File`. I was thinking about re-configuring the validations so that they were methods inside the class they're validating, rather than separate classes. That might make it easer to understand. Something like:
``` ruby
class Bag::File
  validate :object_id_must_match_parent

  def object_id_must_match_parent
    return if object_id == file.parent.basename.to_s
    errors.add(:file, "blah....")
  end

end

app/cho/bag/work.rb
awead: @cam156 I thought about that, but it seemed to over-complicate things. If it is, should there be separate unit tests as well?

app/cho/bag/work_validator.rb
awead: @cam156 A representative isn't a file type. It's an unnamed type of FileSet. The distinction is drawn by looking at all the files in the set. If none of them have an id segment, it's a representative file set, i.e. it's representative of the work. There's an example [here](https://github.com/psu-libraries/cho/wiki/File-Specifications#simple-work-folder-of-manuscript-materials) where there's a file set that has the pdf of all the pages, and separate file sets for each individual page.

spec/cho/bag/package_spec.rb
awead: It's the directory name. I chose a hash structure because it seemed to best reflect the directory structures in https://github.com/psu-libraries/cho/wiki/File-Specifications

spec/cho/bag/work_spec.rb
awead: @cam156 not exactly. They'll be works when they're imported. I was following the examples in the specification, but `workID` might be a lot clearer.

awead:
@cam156 the package is the bag. I needed another noun to reflect what our bags were, otherwise it would be `Bag::Bag`. I'm open to other suggestions.

app/cho/bag/file_validator.rb
cam156: @awead I wonder if this is another opportunity for dry-transactions?

cam156:
@awead paper? 😄 Does it represent a a work at the top level or a List or Works?

app/cho/bag/file_validator.rb
awead: @cam156 we could, but we wouldn't get the `valid?` method, unless we want the transaction to call that. I can definitely see dry-transactions for the actual ingest/import, in which case a valid operation would call `valid?` on the bag and report the errors accordingly. The nice thing about AR validations is you get the `valid?` method and the `errors` object.

carolyncole (approved):

PR #579, opened 2018-06-28
Allow core fields to be overridden

cam156:
## Description
add sort to catalog data dictionary display
add sort to core fields
make sure the order index does not get overriden

Fixes: #437

Why was this necessary?
Allows core field to be overridden without killing the order index.  Not this does not update the order on the show and index views, but does update the order on the edit and new forms.

## Changes

## Checklist

- [ ] i18n enabled
- [ ] adequate test coverage
- [ ] accessible interface

### DIFF
--- a/spec/support/seed_map.rb
+++ b/spec/support/seed_map.rb
@@ -3,4 +3,4 @@
 # This loads the database so that the classes can be set up correctly
 #   Catalog Controller and Work::Submission dynamically build code based
 #   on this seeded data dictionary
-Rails.application.load_seed
+Rails.application.load_seed if ENV['RAILS_ENV'] == 'test'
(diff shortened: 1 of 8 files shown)


spec/support/seed_map.rb
```diff
@@ -3,4 +3,4 @@
 # This loads the database so that the classes can be set up correctly
 #   Catalog Controller and Work::Submission dynamically build code based
 #   on this seeded data dictionary
-Rails.application.load_seed
+Rails.application.load_seed if ENV['RAILS_ENV'] == 'test'
```
awead: @cam156 how would this get called if it wasn't the test environment? `spec/support` files should only get loaded when the test suite runs.

cam156: @awead That was my understanding as well, but when I was debugging I noticed this was getting called every time.  I'm really not certain.

PR #592, opened 2018-07-10
Implement file sets with GUI upload

awead:
## Description

Takes the existing single-file upload process done in the GUI and adds
the file set resource as the intermediary between the work and file.

## Changes

Single files that are uploaded via the GUI are now stored in a file set that is attached to the work resource.

## Checklist

- [x] i18n enabled
- [x] adequate test coverage
- [x] accessible interface


awead:
I've verified this is working locally, and am bypassing the review since @cam156 and I wrote this up together.

Issue #443, opened 2018-03-09
Collection-ID Field on Work Create Page


ntallman:
"Member of collection ids" should have a required label and validation

![image](https://user-images.githubusercontent.com/549591/37231314-ae7071f4-23b8-11e8-9fe9-235783b6312c.png)

awead:
To fix:
* rename "collection" to "member_of_collection_ids" in data dictionary
* create a "system collections" controlled vocabulary
* display HTML5 datalist controlled vocabulary in form
* remove explicit member_of_collection_ids in models, change sets, and forms, and utilize the data dictionary term instead
* update dictionary term to be "required"

awead:
Related to #489

PR #670, opened 2018-09-26
Define resource and change set property types for fields

awead:
## Description

Adds methods to DataDictionary::Field to return the specific Valkyrie::Type for a given field when it is assigned to either a resource or a change set.

Connected to #443

## Changes

* adds `change_set_property_type` and `resource_property_type` methods to DataDictionary::Field
* fixes a latent bug in CsvDryRun where the change sets were returning singular values instead of multiple ones


awead:
Coveralls is a false-positive. Those lines in Import::Work are covered.

Issue #618, opened 2018-07-26
Transaction::Import::ImportFromZip failures


awead:
If Transaction::Import::ImportFromZip fails, it can return either an Import::Bag, if the bag itself extracted correctly but was invalid somehow; or, if there was an unknown error that raised an exception, a string is returned.

This causes an error at line 43 of CsvDryRunResultsPresenter because the bag is a failure, but due to the exception thrown, it's a string object and not an Import::Bag.

Maybe the transaction should always return an Import::Bag, even in the event of an unknown exception?

awead:
Opted for `Rejection` class solution.

PR #681, opened 2018-10-08
Use a Rejection to capture failures from transactions

awead:
## Description

Return a Transaction::Rejection object so we can add failure errors that are interpretable by the presenter.

Connected to #618


awead:
See abe68f0014580c8b933cd349b16ec0e4d8247fce. This wraps the failure in a rejection object that has the message in its `errors`.

Issue #652, opened 2018-09-13
Create file of extracted text and add it to the file set


awead:
Child of #135

Use hydra-derivatives to add extracted text content to a file's file set.

awead:
This work was originally pointed in #135, and was an 8.

PR #699, opened 2018-10-16
Store extracted text on the file set

awead:
## Description

Perform text extraction when uploading files.

Connected to #652

## TODOs

* failures in extraction are ignored for now (see #700 for follow-up)


awead:
@cam156 if you have a chance, I'd like your opinion on this.

Issue #650, opened 2018-09-13
Export Collection Metadata, Including Fileset Metadata


ntallman:
As a metadata librarian,
I would like to be able to export the metadata for all the works (and their filesets) in a collection as a CSV,
So that I can curate, remediate, and otherwise edit the metadata for reimport OR for export for other projects.

Blocked by #582

awead:
Details:
* expand the csv export of a collection to include all the file sets included within each work
* each file set within a work will result in another row in the csv

PR #707, opened 2018-10-22
Include file set metadata when exporting collections

awead:
## Description

Adds additional lines to the csv when exporting collections and works containing file sets.

Connected to #650


awead:
Reopened as #709 because Travis was having issues due to Github outage.

PR #708, opened 2018-10-22
Update transactions to use rejection class

### LINKED ISSUE #643: Rendering Failure messages from operations
Controllers are not able to interpret Failure messages sent from operations. Currently, the messages are just strings, but we might want to put them into the `errors` object and attach it to the change set so that they can be displayed as a flash message, like we are currently doing with other types of errors.

Related to #553

mtribone:
Updates transaction operations to use the Rejection class for errors including their corresponding tests.

## Description

Fixes #643


mtribone:
@awead I wondered about the `Error => e` and didn't mention it. I'll fix those up this morning.

Issue #672, opened 2018-09-26
Using our validators


awead:
We define and assign validators using Validation::Factory, but it's unclear how they can translate into validating properties on the change set. We employ ActiveModel::Validations on change sets, for example:

```
class Resource < Valkyrie::Resource
  property :foo
  validates :foo, with: :bar
end
```

Creating a validator using Validation::Factory.lookup(:bar) doesn't work because the change set is expecting a method named "bar" and we have returned a class.

PR #742, opened 2018-12-05
Implementing an alternative validation option

cam156:
## Description

This utilizes the factory to look up the correct validator
Adds errors to validator to allow for validation feedback

This is an alternative way to run the validation.  It more closely matches what I was thinking about when I put the factory in.

This change adds no new functionality, but I believe may be a better pattern when we start including more validators.

Connected to #672


app/cho/validation/edtf_date.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module Validation
+  class EDTFDate < Base
+    def validate(field_value)
+      @errors = []
```
awead: @cam156 could we set @errors to an empty array in Base so we wouldn't have to do it each time in the subclasses?

spec/cho/validation/edtf_date_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    end
+
+    context 'with a multiple invalid EDTF dates' do
+      let(:dates) { ['asdf', '2'] }
+
+      it 'is invalid' do
+        expect(validation_instance.validate(dates)).to be_falsey
+        expect(validation_instance.errors).to contain_exactly(
+          'Date asdf is not a valid EDTF date',
+                                                                 'Date 2 is not a valid EDTF date'
```
awead: weird spacing issue here. I would have thought Rubocop would fix this for us :(

app/cho/validation/edtf_date.rb
cam156: I was thinking this allowed for validation to be called more than once on an instance.
```
date1 = "blah"
date2 = "2001-02-uu"
validator = Validation::EDTFDate.new
validator.validate(date1)
puts validator.errors # bad date
validator.validate(date2)
puts validator.errors # empty
```

The way the factory returns a cached instance currently I think there would be a bleeding of errors if we do not reset it each time.

Issue #741, opened 2018-12-04
Index thumbnail path


awead:
Add the path to the thumbnail to the solr document. This should be a text field such as "thumbnail_ss" present in file set and work documents. A work's thumbnail is assumed to be the first file set.

PR #746, opened 2018-12-06
Index path to the thumbnail for works and file sets

awead:
## Description

Connected to #741

Adds a NullFileSetRepresentative for works that have no file sets, as
well as adds better test coverage to our existing indexers.


cam156:
@awead did you still want this based on the file not being displaying?

awead:
@cam156 Yes, @mtribone and I are discussing using a custom thumbnail presenter from Blacklight to make the necessary pathing decisions. This has the advantage of keeping the path to the actual file, and the symlinked path to the thumbnail separate. That way, if the location of the symlink changes, or there is a new location for thumbnails, we don't have to re-index solr with the new path information.

Issue #740, opened 2018-12-04
Display thumbnails


awead:
If the file set contains a thumbnail, it should be displayed. Do not display any placeholder icons if there is no thumbnail (this may be changed later).

For works, default the thumbnail to the first file set in the work.

For testing purposes, a bag with an existing thumbnail will be provided.

Related to #741

awead:
Path to the thumbnail will be in the solr document.

TODO:
* update file set show page to display thumbnail
* update work show page to display thumbnail
* update search results page to display thumbnails

PR #748, opened 2018-12-06
Display thumbnails in search results and show views

awead:
## Description

Using Blacklight's features and overriding them where needed, we're displaying thumbnails on search results and show views for works and file sets.

Connected to #740 

## Changes

* creates custom Blacklight presenters
* moves files storage to the public directory for easier linking
* indexes a thumbnail's path in the work and file set
* identifies representative file sets for works

### DIFF
--- a/app/views/catalog/_show_main_content.html.erb
+++ b/app/views/catalog/_show_main_content.html.erb
@@ -1,0 +1,18 @@
+<%# Removes the pagination code from here to the app/views/catalog/show partial for better responsive pages  %>
+<% @page_title = t('blacklight.search.show.title', document_title: document_show_html_title, application_name: application_name).html_safe %>
+<% content_for(:head) { render_link_rel_alternates } %>
+
+<div id="document" class="document <%= render_document_class %>" itemscope itemtype="<%= @document.itemtype %>">
+  <div id="doc_<%= @document.id.to_s.parameterize %>">
+    <%= render_document_partials @document, blacklight_config.view_config(:show).partials %>
+  </div>
+</div>
+
+<% if @document.respond_to?(:export_as_openurl_ctx_kev) %>
+  <!--
+       // COinS, for Zotero among others.
+       // This document_partial_name(@document) business is not quite right,
+       // but has been there for a while.
+  -->
+  <span class="Z3988" title="<%= @document.export_as_openurl_ctx_kev(document_partial_name(@document)) %>"></span>
+<% end %>
--- a/config/initializers/valkyrie.rb
+++ b/config/initializers/valkyrie.rb
@@ -39,7 +39,7 @@ Rails.application.config.to_prepare do
   # Storage Adapters
 
   Valkyrie::StorageAdapter.register(
-    Valkyrie::Storage::Disk.new(base_path: Rails.root.join('tmp', 'files')),
+    Valkyrie::Storage::Disk.new(base_path: Metrics::Repository.storage_directory),
     :disk
   )
 
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@
 .rubocop-http*
 config/database.yml
 config/hydra-ldap.yml
+public/files
--- a/app/blacklight/catalog_controller.rb
+++ b/app/blacklight/catalog_controller.rb
@@ -7,6 +7,8 @@ class CatalogController < ApplicationController
 
   helper LocalHelperBehavior, LayoutHelperBehavior
 
+  Blacklight::IndexPresenter.thumbnail_presenter = ::ThumbnailPresenter
+
   configure_blacklight do |config|
     ## Class for sending and receiving requests from a search index
     # config.repository_class = Blacklight::Solr::Repository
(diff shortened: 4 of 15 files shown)


app/cho/work/submission_indexer.rb
```diff
@@ -21,7 +22,7 @@ def to_solr
       # @note returns the label associated with the work_type_id
       def label_for_id
         work_type = Work::Type.find(Valkyrie::ID.new(resource.work_type_id))
-        return resource.work_type_id if work_type.nil?
+        return resource.work_type_id.to_s if work_type.nil?
```
cam156: @awead is this change needed?

app/views/catalog/_show_main_content.html.erb
```diff
@@ -0,0 +1,17 @@
+<% @page_title = t('blacklight.search.show.title', document_title: document_show_html_title, application_name: application_name).html_safe %>
```
cam156: What is this override providing?

config/initializers/valkyrie.rb
```diff
@@ -39,7 +39,7 @@
   # Storage Adapters
 
   Valkyrie::StorageAdapter.register(
-    Valkyrie::Storage::Disk.new(base_path: Rails.root.join('tmp', 'files')),
+    Valkyrie::Storage::Disk.new(base_path: Metrics::Repository.storage_directory),
```
cam156: This seems odd having our initializer dependant on metrics.

spec/cho/work/submissions/indexer_spec.rb
```diff
@@ -15,32 +15,30 @@
...
-      it { is_expected.to eq(work_type_ssim: "I don't exist", member_of_collection_ssim: nil) }
+      it { is_expected.to include(work_type_ssim: "I don't exist") }
     end
 
     context "when the resource's work type does exist" do
       let(:work_type) { create :work_type, label: 'Indexed Label' }
-      let(:resource) { instance_double('Resource', work_type_id: work_type.id) }
+      let(:resource)  { create(:work, work_type_id: work_type.id) }
 
-      it { is_expected.to eq(work_type_ssim: 'Indexed Label', member_of_collection_ssim: nil) }
```
cam156: Did member_of_collection_id go away for work?  Why is this needed for thumbnails?

app/cho/work/submission.rb
```diff
@@ -34,8 +34,23 @@ def i18n_key
       end
     end
 
+    def file_sets
+      @file_sets ||= file_set_ids.map { |id| Work::FileSet.find(id) }
```
cam156: I think this could get stale.  Aka if file_set_ids gets set on the work the @file_sets should be set to nil

app/cho/work/submission_indexer.rb
awead: It was a hidden bug that was exposed when I refactored the spec test. If the work's type has an id that doesn't exist yet, you'll get a solr field like: `work_type_ssim => #<Valkyrie::ID:0x007fdaa6852ef0 @id="I don't exist">` so we have to `to_s`. That being said, it doesn't make much sense for `work_type_ssim` to be an id if the type hasn't been saved yet or is non-existent. How about just return nil if the type isn't there:
``` ruby
return if work_type.nil?
```

app/views/catalog/_show_main_content.html.erb
awead: Not sure. @mtribone ?

app/cho/work/submission.rb
awead: @cam156 Do you think not memoizing could have performance impacts if we need to re-query for files sets? Or, should be have a reload mechanism that can refresh these? Note, this applies in other areas too and not just file sets, so there's an opportunity to set a pattern here.

config/initializers/valkyrie.rb
awead: @cam156 Yeah, agreed. My intent is to refactor that out in another ticket. I didn't want it get too far afield. I can link a new ticket to this so we don't loose it.

spec/cho/work/submissions/indexer_spec.rb
awead: @cam156 This was part to the refactor. I stopped using instance_double and started using the factory. This is also just testing work type, so I'm only verifying that work_type_ssim is set. In the factory, it already has a collection, so it's not nil anymore, but just checking for work type keeps the test more focused.

app/views/catalog/_show_main_content.html.erb
mtribone: Ah! Not needed anymore as the thumbs and layout have been moved to another partial. Thanks!

mtribone: oh wait. It is to remove the pagination from the top when we are viewing works and filesets. It is needed.

mtribone: Nope. This feature was specifically asked for in the user stories. Putting it back in, but in another partial.

https://github.com/psu-libraries/cho/issues/124

config/initializers/valkyrie.rb
awead: Here's the ticket: #750

app/cho/work/submission_indexer.rb
awead: This was resolved in a previous commit.

Issue #733, opened 2018-11-15
Form for adding creators to works or collections


awead:
Builds the necessary html for entering creators and roles in a work or collection. This also includes the code required to validate the role with the Agent and return any errors if either is invalid. The Agent controlled vocabulary is already done, but the roles should be specified and validated here as MARC relators.

MARC relators can be a simple datalist pulled from http://id.loc.gov/vocabulary/relators.html. A json file might be best here. 

Child of #87 
Blocked by #732 
[Related: #191  NT]

awead:
TODO
* create a validator that checks that the creator and roles exist
* use the field type from #732 to create the necessary UI elements to display dropdowns
* agents and roles HTML5 datalist elements

PR #761, opened 2018-12-13
Form for adding creators to works or collections

awead:
## Description

Enables adding a creator to a work or collection via the GUI.

Connected to #733 

## Changes

* includes a MockRDF class for mocking large RDF vocabularies across the entire test suite

### DIFF
--- a/app/cho/data_dictionary/with_index_type.rb
+++ b/app/cho/data_dictionary/with_index_type.rb
@@ -3,7 +3,7 @@
 module DataDictionary::WithIndexType
   extend ActiveSupport::Concern
 
-  IndexTypes = Valkyrie::Types::String.enum('facet', 'no_facet', 'date')
+  IndexTypes = Valkyrie::Types::String.enum('facet', 'no_facet', 'date', 'linked_field')
 
   included do
     attribute :index_type, IndexTypes
@@ -31,5 +31,13 @@ module DataDictionary::WithIndexType
 
   def date?
     index_type == 'date'
+  end
+
+  def linked_field!
+    self.index_type = 'linked_field'
+  end
+
+  def linked_field?
+    index_type == 'linked_field'
   end
 end
--- a/app/cho/indexing/creator.rb
+++ b/app/cho/indexing/creator.rb
@@ -1,0 +1,60 @@
+# frozen_string_literal: true
+
+module Indexing
+  class Creator
+    attr_reader :hash_values, :solr_field
+
+    def initialize(values:, solr_field:)
+      @hash_values = values.map { |hash| HashCreator.new(hash) }
+      @solr_field = solr_field
+    end
+
+    def to_solr
+      {
+        solr_field => hash_values.map(&:display_name).compact,
+        'creator_role_tesim' => hash_values.map(&:role).compact,
+        'creator_role_ssim' => hash_values.map(&:role).compact,
+        'creator_agent_tesim' => hash_values.map(&:agent_display).compact,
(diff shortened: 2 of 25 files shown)


app/cho/indexing/linked_fields.rb
```diff
@@ -0,0 +1,75 @@
...
+        @linked_fields ||= Schema::MetadataField.where(index_type: 'linked_field').select do |field|
+          resource.attributes[field.label.to_sym].present?
+        end.to_a
+      end
+
+      def solr_fields(field)
+        creators = resource.attributes[field.label.to_sym].map { |hash| HashCreator.new(hash) }
+        {
+          field.solr_field => creators.map(&:display_name).compact,
+          'creator_role_tesim' => creators.map(&:role).compact,
```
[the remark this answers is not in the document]
awead: @cam156 Yeah, I couldn't figure out a way to make this more versatile for different kinds of linked fields. I think I just gave up here 😄 Any suggestions?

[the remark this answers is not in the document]
cam156: @awead I was thinking the linked field could generate its solr data, or have an indexer for each field type.

awead:
@cam156 I moved the creator solr field stuff out to is own class. The LinkedField class will now try to find a class based on the field's label. It's not a great solution, but I think it's getting us closer towards the right overall solution for indexing. LinkedField and Dates start to look more similar. The field class might be able to call these indexing classes as well, which could result in a more general indexing class that would work with any field. There's definitely more packed in here, but it needs a separate ticket and some more thought.

app/cho/data_dictionary/with_index_type.rb
```diff
@@ -32,4 +32,12 @@ def date!
...
     index_type == 'date'
   end
+
+  def linked_field!
+    self.index_type = 'linked_field'
+  end
+
+  def linked_field?
+    index_type == 'linked_field'
+  end
```
mtribone: What is `linked_field!` ? I understand `linked_field?` is checking to see if this is a linked field, but I don't get what the "!" does for linked_fields.

awead: @mtribone the `!` in Ruby means it would alter the existing field record to be linked field index type. If I call `field.linked_field!` it sets its `index_type` to be "linked_field".

mtribone: Okay. So the difference between the two methods `!` and `?` in this file is that the `!` transforms _existing_ fields and the `?` is seeing if _it is possibly_ a linked field and if so assigning its index type.

awead: @mtribone that's right. The `!` in ruby means, "I am changing this object"

app/cho/indexing/creator.rb
```diff
@@ -0,0 +1,58 @@
...
+    def initialize(values:, solr_field:)
+      @hash_values = values.map { |hash| HashCreator.new(hash) }
+      @solr_field = solr_field
+    end
+
+    def to_solr
+      {
+        solr_field => hash_values.map(&:display_name).compact,
+        'creator_role_tesim' => hash_values.map(&:role).compact,
+        'creator_role_ssim' => hash_values.map(&:role).compact,
```
cam156: Should we be including role id  and agent id in there so that the string is not ambiguous?

cam156: @awead when someone puls the record back and tries to do an action on it.  I was thinking something simple like two more lines
```
'creator_role_id_ssim' => role_id,
'creator_agent_id_ssim' => agent_id,
```

PR #767, opened 2019-01-08
Delete contained works from collections

### LINKED ISSUE #509: Confirmation when deleting collections and works
When I choose to delete a collection
I should see a confirmation page showing all of the contained works that will be deleted
And the collection as well as all of its works and files should be deleted.

awead:
## Description

Presents a confirmation message with a list of contained items when deleting works or collections and includes the counts of those items in the success message.

Fixes #509

### DIFF
--- a/app/cho/batch/delete_presenter.rb
+++ b/app/cho/batch/delete_presenter.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+module Batch
+  class DeletePresenter
+    attr_reader :resource, :children
+
+    delegate :id, :title, to: :resource
+
+    def initialize(resource)
+      @resource = resource
+      @children = (titles_for(:members) + titles_for(:file_sets) + titles_for(:files)).flatten
+    end
+
+    def confirmation_title
+      I18n.t('cho.batch.delete.title', title: title.first, count: children.count)
+    end
+
+    private
+
+      def titles_for(type)
+        contained_resources(type).map do |member|
+          if type == :files
+            "File: #{member.original_filename}"
+          else
+            ["#{member.model_name.human}: #{member.title.first}"] + self.class.new(member).children
+          end
+        end
+      end
+
+      def contained_resources(type)
+        return [] unless resource.try(type)
+        resource.send(type)
+      end
+  end
+end
--- a/spec/cho/batch/delete_controller_spec.rb
+++ b/spec/cho/batch/delete_controller_spec.rb
@@ -27,7 +27,7 @@ RSpec.describe Batch::DeleteController, type: :controller do
         delete :destroy, params: { delete: { ids: [work.to_param, collection.to_param] } }
       }.to change { metadata_adapter.query_service.find_all.to_a.count }.by(-2)
       expect(flash[:success]).to eq(
-        "You have successfully deleted the following items: #{[work.title.first, collection.title.first].join(', ')}"
+        I18n.t('cho.batch.delete.success', list: 'Sample Generic Work (0 items), Archival Collection (0 items)')
       )
     end
   end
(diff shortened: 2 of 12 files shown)


app/cho/batch/delete_presenter.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    delegate :id, :title, to: :resource
+
+    def initialize(resource)
+      @resource = resource
+      @children = member_titles.flatten + file_set_titles.flatten + file_titles.flatten
+    end
+
+    def confirmation_title
+      "#{title.first} (#{children.count} items)"
```
cam156: Items was confusing to me when I first looked at it.  I wonder if this should be I18N so it can be changed later.

app/cho/batch/delete_presenter.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+
+    private
+
+      def member_titles
+        return [] unless resource.try(:members)
+        resource.members.map do |member|
+          ["Work: #{member.title.first}"] + self.class.new(member).children
```
cam156: Why are the children added here?  Is this to show works that include files sets?

app/cho/batch/delete_presenter.rb
```diff
@@ -0,0 +1,41 @@
...
+        end
+      end
+
+      def file_set_titles
+        return [] unless resource.try(:file_sets)
+        resource.file_sets.map do |file_set|
+          ["FileSet: #{file_set.title.first}"] + self.class.new(file_set).children
```
cam156: Does the children here show all the different files?

spec/cho/batch/delete_controller_spec.rb
```diff
@@ -27,7 +27,8 @@
         delete :destroy, params: { delete: { ids: [work.to_param, collection.to_param] } }
       }.to change { metadata_adapter.query_service.find_all.to_a.count }.by(-2)
       expect(flash[:success]).to eq(
-        "You have successfully deleted the following items: #{[work.title.first, collection.title.first].join(', ')}"
+        'You have successfully deleted the following items: '\
```
cam156: Should this be I18N?

app/cho/batch/delete_presenter.rb
awead: Yes, for works and file sets we want to recurse back into the presenter to get their file sets and files.

app/cho/batch/delete_presenter.rb
awead: @cam156 Yes. I have an update that's hopefully going to make this clearer.

spec/cho/batch/delete_controller_spec.rb
awead: @cam156 It already is, but it takes a `list` variable. Should I check I18n directly?

app/cho/batch/delete_presenter.rb
```diff
@@ -0,0 +1,35 @@
...
+    end
+
+    private
+
+      def titles_for(type)
+        contained_resources(type).map do |member|
+          if type == :files
+            "File: #{member.original_filename}"
+          else
+            ["#{member.model_name.human}: #{member.title.first}"] + self.class.new(member).children
```
cam156: @awead I like this.

spec/cho/batch/delete_controller_spec.rb
cam156: @awead I was thinking about the "You have successfully deleted the following items" portion

app/cho/batch/delete_presenter.rb
cam156: Can collections be members of collections?  I think the answer is no, but just checking.

awead: @cam156 No, because it's an inverse relationship. Only works assert their membership to a collection.

awead:
@cam156 OK, this should be fixed up now. Thanks for the comments!

Issue #633, opened 2018-08-30
Edit forms for file sets


awead:
Create an edit page for editing the metadata of a file set. This should not include any abilities for adding or deleting files.

We can use a schema for the file set's fields, however core fields are automatically included and fields such as "collection" do not apply. There may need to be a way for schemas to exclude specific fields.

Related to #63

awead:
File set show page need to resemble more closely the work page:
*edit links
* a link back to the parent work
* some indication that this is a file set

The actual edit page can be the same as the work edit page.

PR #770, opened 2019-01-10
Fileset Show Page - Part 1

mtribone:
## Description

Gets work title using inverse references and a query service via the Solr document because it is "easier" this way than trying to create a two directional relationship between the fileset and the work. Checks to see if there is a thumbnail since we are using the same show partial for filesets and works and displays one or two columns.

Connected to #633

### DIFF
--- a/app/views/catalog/show.html.erb
+++ b/app/views/catalog/show.html.erb
@@ -10,12 +10,18 @@
 <%# Use Bootstrap columns to position the thumbnail on the left and the content partials on the right %>
 <div class="container mt-4">
   <div class="row">
-    <div class="col-xs-12 col-md-4">
-      <%= render 'thumbnail', document: @document, document_counter: 0 %>
-    </div>
-    <div class="col-xs-12 col-md-8">
-      <%= render_document_main_content_partial %>
-    </div>
+    <% if presenter(@document).thumbnail.exists? %>
+      <div class="col-xs-12 col-md-4">
+        <%= render 'thumbnail', document: @document, document_counter: 0 %>
+      </div>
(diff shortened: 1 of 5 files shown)


app/views/catalog/show.html.erb
```diff
@@ -10,12 +10,24 @@
...
-    <div class="col-xs-12 col-md-8">
-      <%= render_document_main_content_partial %>
-    </div>
+    <% if presenter(@document).thumbnail.exists? %>
+      <div class="col-xs-12 col-md-4">
+        <%= render 'thumbnail', document: @document, document_counter: 0 %>
+      </div>
+      <div class="col-xs-12 col-md-8">
+        <% if @document.internal_resource == Work::FileSet %>
+          <p>Fileset | Part of: <%= link_to @document.work_title, solr_document_path(@document.work) %></p>
```
awead: @mtribone Can we put the "Fileset" vs. "Fileset | Part of:" logic in _show_work_fileset_html.erb instead?

awead: @mtribone Sorry, looks like this has been updated to just "Fileset | Part of:" but I think it could go under the file set's own show partial.

mtribone: screen shot 2019-01-10 at 10 08 02 am

awead:
@mtribone Looks good. Do you want to continue with this PR, or merge and start a new one for the edit functions?

mtribone:
@awead Yes. I'd like to start a new PR for the edit functions. Thanks!

mtribone:
screen shot 2019-01-10 at 10 08 02 am

Issue #773, opened 2019-01-11
URL-encode all data in the forms


awead:
This would avoid errors from Apache if it thinks something nefarious is going on when we send unencoded RDF urls.

ntallman:
No longer needed to avoid the original error, but probably not a bad idea. Putting into the backlog for future consideration.

PR #776, opened 2019-01-11
Changing to use path instead of entire URL to avoid apache error

cam156:
refs #773

@awead this is a simple option to avoid the apache issue.  It may also make the UI look cleaner by removing the http junk from the front.
screen shot 2019-01-11 at 3 45 32 pm

I have yet to deploy this to QA to see if it solves the problem, but I believe it would.

Connected to #773


cam156:
@awead it does.  Did we want to try deploying it before we merge?

awead:
@cam156 Jim fixed the issue with mod_security, so we can now push links in our forms. QA and staging servers are set to report links that are found and then rules are developed to allow them. These rules are then deployed to production so that only "known" links are let through. This is how Scholarsphere is set up as well.

In the future, if we send other links in our forms, Apache will report the problem (but allow them) and we'll be able to update the whitelisted links prior to deploying to production.

PR #787, opened 2019-01-23
Updating to Blacklight 7.0.1

awead:
## Description

Update Blacklight to 7.0.1.

### DIFF
--- a/app/views/catalog/_document_members.html.erb
+++ b/app/views/catalog/_document_members.html.erb
@@ -8,10 +8,9 @@
     </thead>
     <tbody>
       <% documents.each do |document| %>
-        <% member_presenter = show_presenter(document) %>
         <tr>
           <td><%= link_to_document document, document_show_link_field(document) %></td>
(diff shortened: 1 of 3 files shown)


app/views/catalog/_document_members.html.erb
```diff
@@ -11,7 +11,7 @@
         <% member_presenter = show_presenter(document) %>
         <tr>
           <td><%= link_to_document document, document_show_link_field(document) %></td>
-          <td><%= member_presenter.field_value('created_at_dtsi') %></td>
+          <td><%= document['created_at_dtsi'] %></td>
```
cam156: This seems like a strange change to no longer utilize the presenter

awead: @cam156 I don't think we should have been using the presenter in this way to start with, so I took it out. The method signature changed for `field_value` in the new Blacklight version, and we were getting an error here. Perhaps we were using it in a way not intended?

We aren't doing any special display processing with the created date, so the simplest way seems to be just to use the solr document.

PR #852, opened 2019-03-06
Adds the ability to update file sets through the UI

mtribone:
## Description

Updates the tests and partials for the ability to update and edit file sets using the UI

Connected to #633 

Co-authored-by: Adam Wead <amsterdamos@gmail.com>


app/cho/work/file_set.rb
```diff
@@ -54,4 +55,19 @@ def text_source
...
+  # @note Use a custom Name class to preserve default behaviors while overriding others
+  class Name < ActiveModel::Name
+    def i18n_key
+      'file_set'
+    end
+
+    def human
+      'File Set'
+    end
+  end
```
awead: @mtribone we should add a test for this. Copy from `submission_spec.rb`:

``` ruby
  describe '::model_name' do
    context 'with a custom i18n key' do
      subject { described_class.model_name }

      its(:i18n_key) { is_expected.to eq('work') }
      its(:human) { is_expected.to eq('Work') }
    end
  end
```

app/cho/work/file_set_change_set.rb
```diff
@@ -18,6 +21,28 @@ def initialize(*args)
       super(*args)
     end
 
+    # @param [ActionView::Helpers::FormBuilder] form
+    # @return [Array<Schema::InputField>]
+    def input_fields(form)
+      form_fields.map { |field| Schema::InputField.new(form: form, metadata_field: field) }
+    end
```
awead: @mtribone add a test for this. There's one in `submission_spec.rb` that you can copy right out.

app/cho/work/file_set_change_set.rb
```diff
@@ -18,6 +21,28 @@ def initialize(*args)
...
+
+    def form_path
+      if model.persisted?
+        url_helpers.file_set_path(model)
+      else
+        url_helpers.file_sets_path
+      end
+    end
```
awead: @mtribone we can remove this method. We copied it from the submission change set. Because we aren't creating new file sets, only editing existing ones, we'll never hit the else case. The model is always persisted. But you'll need to change `app/views/work/file_sets/_form.html.erb` to:

    <%= form_for(file_set_form.model) do |form| %>

app/cho/work/file_set_change_set.rb
```diff
@@ -18,6 +21,28 @@ def initialize(*args)
...
+
+    def submit_text
+      if model.persisted?
+        I18n.t('cho.file_set.edit.submit')
+      else
+        I18n.t('cho.file_set.new.submit')
+      end
+    end
```
awead: @mtribone Same thing here. Because we're not creating new file sets, we won't need this conditional. Remove this method and then edit `app/views/work/file_sets/_form.html.erb` to use an i18n key:

    <%= form.submit t('cho.file_set.edit.submit') %>

app/views/catalog/show.html.erb
```diff
@@ -25,9 +25,8 @@
   </div>
 </div>
 
-<% unless @document.internal_resource == Work::FileSet %>
-  <%= link_to 'Edit', edit_cho_resource_path %>
-<% end %>
+<%= link_to 'Edit', edit_cho_resource_path %>
```
awead: @mtribone add i18n to this.

app/cho/work/file_set.rb
mtribone: @awead can this `describe` just go in the `submission_spec`, but update it to represent file sets? We have an `indexer_spec` in file_set, but not sure if it really goes there. Or just add an `update_spec` to cover this and the `input_fields` from the change set?

awead: @mtribone The example above was taken from `submission_spec.rb`. Copy it and put it in `spec/cho/work/file_set_spec.rb`. You'll need to change the expectations to match what they are for file sets.

mtribone: ah. I was  only looking in `/work/fileset` originally.

Issue #649, opened 2018-09-13
Download Files


ntallman:
As a user,
I would like to be able to click a link and download files attached to a fileset attached to a work,
So that I can save the file/work to my device.

Simplest implementation presents all files with a download link, though this is not the final desired solution.

ntallman:
This might end up being accomplished in #63.

ntallman:
If a fileset contains a _preservation_redacted file, the _preseration file cannot be downloadable/viewable by the user.

awead:
TODO:
* add download links for every file in the file set to the file set show page
  * if a redacted preservation file is present, this is downloaded and _not_ the actual preservation file
* add a download link to each file set in the work show page
  * link downloads the service file, if present, or preservation the file

awead:
* Add options for all the other use types except redacted preservation: front, back, admin, text, and thumbnail.
* Remove the download links on the work landing page.
* From the file set landing page, make the title of the file the link

PR #859, opened 2019-03-12
DownloadsController for files

awead:
## Description

Downloads a file from a file set based on parameters in the url and what file types are present in the file set.

Connected to #649 

## Changes

* `Respository::DownloadsController` to return the downloaded content
* `Repository::Download` to determine what the downloadable file is with a given file set
* additional file set factory traits for the different types to files


spec/cho/repository/downloads_controller_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+      end
+    end
+
+    context 'when requesting the preservation file when a redacted one is present' do
+      let(:file_set) { create(:file_set, :with_redacted_preservation_file) }
+
+      specify do
+        get :download, params: { id: file_set.id, use_type: 'PreservationMasterFile' }
+        expect(response).to be_success
+        expect(response.body).to eq('Hello World!')
```
cam156: @awead since these files all return the same content so your code could be doing nothing and the tests would still pass.  It would be better of the redacted text was different from the preservation, was different form the access, so you could tell for sure you were getting different data.

awead: @cam156 there are no other files in the file set, so the content would be nil if it wasn't finding the right file.

awead: @cam156 but we could have a factory that creates text files with arbitrary content.

cam156: @awead ok, It is not necessary to create a random factory. It just seems a bit weird to have all the content the same...

awead: @cam156 yeah, I'll put in a follow-up and see what you think. It would be handy for other situations when need a bit more than just "Hello World!"

PR #864, opened 2019-03-12
Rake task to batch load works from csv

awead:
## Description

Reads all the csv files from the network ingest directory, with their corresponding zip files, and imports them. If needed, it will created a collection with minimal metadata.

This is largely throwaway code, but would be useful to have around for the duration of the MVP cycle where we could reload a bunch of csvs and zips with a single rake command.


lib/tasks/ingest.rake
```diff
@@ -0,0 +1,54 @@
...
+        title: "Collection for #{id}",
+        alternate_ids: [id],
+        visibility: 'public',
+        workflow: 'default'
+      )
+      change_set_persister.persister.save(resource: collection)
+    end
+
+    def collection_exists?(id)
+      change_set_persister.query_service.find_by_alternate_identifier(alternate_identifier: id)
```
cam156: @awead is it possible that the id is an actual collection id, not just an alternate?

awead: Yes, if it was an update CSV that was exported from an existing collection, but these are all intended as new imports.

Issue #720, opened 2018-11-13
Re-Index Facets on Change to "Index Type" in Data Dictionary


ntallman:
When a user edits a field in the data dictionary to change the "Index Type" property to "facet' from "no_facet", and the metadata already contains values for that field, Solr must re-index to populate the facets. This should happen automatically when the change is saved.

Related to #459
Related to #689

awead:
TODO:
* edit DD to change rights or subject from `no_facet` to `facet` and see if that results in a facet appearing
* correct the DD/CatalogController to fix if the above is not the case

awead:
Rights statements are being faceted, but dates are not. We need to decide how we want to facet dates in general.

awead:
Changed date_created to an edtf date.

PR #878, opened 2019-03-13
Data dictionary updates to date_created and rights_statement

ntallman:
## Description
Updated the index_type of two data dictionary fields. One to make it EDTF, the other to test turning on facets.

Connected to #720

## Changes
* Changed date_create to have an index_type of date 
* Changed rights_statement to have an index_type of facet


awead:
@ntallman we could merge this, but there really isn't any assurance that it won't break anything because it's not changing anything in the test environment. I'll try testing this out locally in development and see what happens.

Issue #791, opened 2019-01-25
Visibility (access level) for resources


ntallman:
As a collection curator,
I want to be able to restrict access to some works,
So that I can respect donor wishes and obey laws.

Done looks like an implementation of the following on the work edit forms:

## Access Level
  * Public Access (anyone who can get to CHO)
  * PSU Access (WebAccess authenticated users)
  * Restricted Access (user or group of users)

Default is choice on forms is Public Access. This should be in the data dictionary as locally controlled vocab so it can be set via batch process.

awead:
Related tickets:
* Enable setting of access levels in CSV - #976

PR #879, opened 2019-03-13
Update UI labels

ntallman:
## Description
Updating various UI labels.

Connected to #791

## Changes
* Changing all references of work to resource
* Changing visibility to access level, clarifying label text for options
* Fixing capitalization in a few places
* Changing destroy to delete
* Changing batch delete labels from items to resources or in some cases dropping the object-type to avoid ambiguity.


config/locales/cho.yml
```diff
@@ -11,85 +11,85 @@ en:
...
       metadata: "Basic Metadata"
       workflow: "Workflow"
-      visibility: "Visibility"
+      visibility: "Access Level"
       file_selection: "File Selection"
     field_label:
       # For fields that are not in the data dictionary
       collection_type: "Collection Type"
-      work_type: "Work Type"
+      work_type: "Resource Type"
```
mtribone: It looks like the change from 'Work Type' to 'Resource Type' has broken a few tests.

config/locales/cho.yml
```diff
@@ -11,85 +11,85 @@ en:
...
+        destroy: "Delete"
         confirm: "Are you sure you want to delete this agent?"
         given_name: "Given Name"
         surname: "Surname"
     csv:
       agent:
         create:
           heading: 'CSV Agent Import'
-          description: 'Import a csv that creates new agents.'
+          description: 'Import a CSV that creates new agents.'
```
mtribone: This change to all caps has broken rspec ./spec/cho/agent/import/update_spec.rb:15

config/locales/cho.yml
```diff
@@ -11,85 +11,85 @@ en:
...
           submit: 'Preview Import'
         update:
           heading: 'CSV Agent Import'
-          description: 'Import a csv that updates existing agents.'
+          description: 'Import a CSV that updates existing agents.'
```
mtribone: Same as above. Good handful of tests that need this change for the all caps CSV.

Issue #686, opened 2018-10-12
Batch Metadata Update (Including filesets) 


ntallman:
As a metadata librarian,
I would like to be able to upload a CSV that has updated metadata for existing works (and filesets) and have that metadata replace the stored values,
So that I can easily update metadata in batch without having to edit each record.

Now that we have added filesets, the import dry run messaging needs to be updated to include success and errors related to filesets. We also need to make the update process change the metadata using the new values provided in the CSV. This should be successful with either a csv export from the collection or the work.

awead:
Not clear on what needs to be done here.

awead:
TODO:
* update the file set metadata when uploading a csv with new values
* display any errors or changes that will occur to file sets in the dry run

awead:
@ntallman RE: last point about missing alternate ids. If the csv has rows for file sets that are missing alternate ids, or even rows that missing completely, the application will still import all the file sets in the bag and attach them to the work. The titles will be their file set identifiers.

Did you want a true warning, meaning that the import would proceed if the ids were missing, or do you want the alt. ids required? If you make alternate identifiers required for file sets, then it would stipulate that all file sets _must_ have metadata rows in the csv. Currently, it's a bit relaxed in that you don't have to provide metadata for all the file sets and CHO just imports them as they are in the bag.

awead:
One other point, it looks like you can omit alt. ids for works, which seems to cause problems. So we ought to make those required in the csv import.

PR #884, opened 2019-03-14
CSV metadata for works and file sets

awead:
## Description

Corrects errors and inconsistencies when creating or updated metadata in file sets via the csv import process.

Connected to #686

### DIFF
--- a/app/cho/work/import/csv_dry_run_results_presenter.rb
+++ b/app/cho/work/import/csv_dry_run_results_presenter.rb
@@ -15,7 +15,7 @@ module Work
       end
 
       def change_set_list
-        dry_run.results
+        dry_run.results + dry_run.results.map(&:file_set_hashes).flatten
       end
 
       def invalid_rows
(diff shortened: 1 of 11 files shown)


app/cho/work/import/csv_dry_run_results_presenter.rb
```diff
@@ -15,7 +15,7 @@ def initialize(dry_run)
       end
 
       def change_set_list
-        dry_run.results
+        dry_run.results + dry_run.results.map(&:file_set_hashes).flatten
```
cam156: @awead I'm a bit confused about this line.  Why are you adjusting the results here?  Seems like you might get the results twice because the results contain the file_set_hashes.

awead: @cam156 `dry_run.results` are just the works and not the file sets. We can map over each work and get its file sets in order to display them in the dry run.

PR #900, opened 2019-03-20
Download files from the file set page only

awead:
## Description

Adds the text, extracted text, thumbnail, front, and back image options for downloads. This includes a substantial refactor to centralize the location of all the use type predicates and the definition of their methods. These methods are now meta-programmed into file sets.

Connected to #649 

## Changes

* adds remaining file use type downloads
* moves all file use uris to `Vocab::FileUse`
* creates `Repository::FileUse` to manage the methods that result from the different use types


app/views/catalog/_show_work_fileset.html.erb
```diff
@@ -8,8 +8,7 @@
   <ul>
     <%- @document.files.each do |file| %>
       <li>
-        <%= file.original_filename %>
-        <%= link_to 'Download',
+        <%= link_to file.original_filename,
```
cam156: @awead this should have some aria description for screen readers.https://www.w3.org/WAI/GL/wiki/Using_aria-label_for_link_purpose

awead:
@cam156 we're subclassing the the PCDM use types in our own vocabulary. I wanted to put all the use types in one vocabulary class so we could make it strict, and it would raise errors if we tried to specify one that didn't exist. I'm not 100% sure that's how these RDF::Vocab classes are meant to be used, so if we need to revisit this later, we can.

awead:
@cam156 I've added an aria label to the download link.

Issue #736, opened 2018-11-15
Validate batch import using agents


awead:
When batch creating works via CSV, verify that non-existent creators are invalidated at the dry-run.

This is done when:
* I can enter "Jane Smith" in the creator field and have the resource created when Jane Smith exists as an Agent
* The dry-run reports errors when I give a creator that doesn't exist as an Agent

Child of #87

awead:
TODO
* dry run will need to match the entered creator with the Agent, so "Jane Smith" needs to map to given name of "Jane" and surname of "Smith"
* can build on the validator built in #733

awead:
@ntallman this will also need to include the role as well, I'm assuming. In the CSV, how should we include all that information? We could do semi-colon separated, ex:

`John Lennon, vocals; Paul McCartney, bass; George Harrison, guitar; Ringo Starr, drums`

This gets a bit more difficult if you have suffixes in the name:

`Frank Sinatra, vocals; Sammy Davis, Jr., vocals`

There'll also be a challenge in matching creators. We'd need to determine first and last name from the string, and this can be problematic. The dry-run will help with that, but we may need to experiment with some different strategies.

ntallman:
I'm consulting on the first part of this. Most best practices I can find advise against mixing information in one field, which might mean we need to do something like Creator and Creator Role; which could be repeated Creator 1, Creator Role 1; Creator 2, Creator Role 2. But that's horrible inellegant, hoping to find a better option. We did something like this for adding files to works via CSV in Cincinnati.

Your second problem is an epic all of it's own and may be out of scope for this ticket. Agent disambuguation has plagued metadata folks forever. I think this is one of the biggest reasons why @ruthtillman is in favor of pulling in Questioning Authority, which I think helps, but I'm not totally sure because I've mostly heard about it being used in the context of forms, not CSV import.

awead:
@ntallman I talked about this with @cam156 and she suggested using separators, which we currently are using in fields like subject. Ex:

`subject1|subject2|subject3`

We could employ a second separator, ex:

`Doe, John||author|Doe, Jane||illustrator`

The double pipe (vertical bar) separates the name from the role, and the single pipe separates each creator/role unit. As far as name disambiguation goes, we can split last name and first name on the comma and CHO is going to match against both those fields. Since we're requiring that the agent resource be present in the system, we should be able to avoid the messy name disambiguation problems from the get-go; however, the challenge is getting a good "interface" in the CSV so that folks can enter names correctly.

[...]

PR #903, opened 2019-03-20
Allow creators with no roles to index correctly

awead:
## Description

Corrects a bug with creators that have no role specified in their csv file.

Connected to #736


spec/cho/work/import/create_spec.rb
```diff
@@ -31,7 +31,7 @@
         creator: [
           "#{agent1.display_name}#{CsvParsing::SUBVALUE_SEPARATOR}bsl",
           "#{agent2.display_name}#{CsvParsing::SUBVALUE_SEPARATOR}cli",
-          "#{agent3.display_name}#{CsvParsing::SUBVALUE_SEPARATOR}bsl"
+          agent3.display_name.to_s
```
cam156: @awead from what I can tell the creator information in the csv is not being checked in the test, otherwise it would have changed when you changed the input.  It would be good to make the test more robust to actually verify the creator.  I realize this should have been done earlier, but it would be good to fix it now.

awead:
@cam156 I added some additional tests to check that the creator name is what it's supposed to be.

Issue #172, opened 2017-07-10
PSU User Role


ntallman:
As a Penn State community member,
I would like to login with my normal credentials,
So that I don't need a separate account to save items in CHO.

Child of #798

ntallman:
Shibboleth not cosign

cam156:
Need to switch out our devise modules to use Shibboleth

ntallman:
Enhancement to #350 that allows additioanl role -- logged in user for things like saving citations and bulk download.

ntallman:
The minimal implementation may use LDAP and psu_dir gem.

The ideal implementation would use Shibboleth/OneId. This would give us the most flexibility in terms of granular permissions later on. Would need to wall off Friends of PSU, Hershey Medical, and Penn College from accessing restricted collections. (Unless stakeholders sign off.) Would also enable us to achieve #181 and #195. ITECH is looking at using OpenAthens for Shibboleth/OneId.

If implementing using the second option, this ticket should be repointed.

ntallman:
Related: #798

awead:
TODO:
* use Cancancan gem to determine access level (admin vs. non-admin)
* restrict access to menu options: manage, create work and collection
* read-only access to everything else

PR #905, opened 2019-03-21
Opened application to Penn State users

rschenk:
## Description

All Penn State users can now access application, but not manage resources.

Connected to #172 

## Changes

* Create/edit/delete navigation links are hidden where appropriate
* Controllers are locked down, and will 403 if an unauthorized user should visit a route

### DIFF
--- a/app/application/application_controller.rb
+++ b/app/application/application_controller.rb
@@ -8,11 +8,12 @@ class ApplicationController < ActionController::Base
   layout 'blacklight'
 
   protect_from_forgery with: :exception
-  before_action :authenticate_user!, :allow_admins
+  before_action :authenticate_user!
+  authorize_resource class: false
 
-  def allow_admins
-    return if current_user.admin?
-
-    render file: 'public/403.html', status: 403, layout: false
+  rescue_from CanCan::AccessDenied do |_exception|
+    render file: Rails.root.join('public', '403.html'),
+           status: 403,
+           layout: false
   end
 end
--- a/app/devise/ability.rb
+++ b/app/devise/ability.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class Ability
+  include CanCan::Ability
+
+  def initialize(user)
+    can :read, :all
+    can :manage, %i[bookmark catalog download search_history]
+
+    can :manage, :all if user&.admin?
+  end
+end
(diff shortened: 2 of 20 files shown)


app/application/application_controller.rb
```diff
@@ -8,11 +8,12 @@ class ApplicationController < ActionController::Base
...
   protect_from_forgery with: :exception
-  before_action :authenticate_user!, :allow_admins
+  before_action :authenticate_user!
+  authorize_resource class: false
 
-  def allow_admins
-    return if current_user.admin?
-
-    render file: 'public/403.html', status: 403, layout: false
+  rescue_from CanCan::AccessDenied do |_exception|
```
awead: @rschenk Is there a way to raise this exception, but without bring the application to a halt? I'm thinking we'd want to know of there were some access denied issues going on, instead of just swallowing it here. We use NewRelic, which as the ability to track error stats. but it's looking for the error in logs. So can we still log/raise the exception and move on? I'm also open to other suggestions as to how to handle this because it will come up in _lots_ of other places.

spec/factories/users.rb
```diff
@@ -22,5 +22,11 @@
       group_list { 'umg/up.libraries.cho-admin' }
       groups_last_update { Time.now }
     end
+
+    factory :psu_user do
+      # This is a made-pu group to contrast with an actual admin group
```
awead: @rschenk oops. I actually typoed this one. Can you reword to "This is a fake group ..."

awead:
This is currently in QA, but we need some non-admin users to verify it works as expected.

app/application/application_controller.rb
awead: I found this https://docs.newrelic.com/docs/agents/ruby-agent/api-guides/sending-handled-errors-new-relic

Something to talk about at standup tomorrow.

awead: We'll use NewRelic's methods to track errors, but extract this out to its own service class for easier management.

app/devise/ability.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class Ability
+  include CanCan::Ability
+
+  def initialize(user)
+    can :read, :all
+    can :manage, %i[bookmark catalog download]
```
awead: @rschenk we'll need to `search_history` as well. This is the action in the log that is getting denied:

`Started DELETE "/search_history/clear`

Other than that, everything looks ok.

Issue #839, opened 2019-02-28
Indexing Bugs


ntallman:
Filenames and identifiers (alternate_ids) don't seem to be indexing anymore.

All of the folloing should but do not [return this work](https://cho-qa.libraries.psu.edu/catalog/ad18ffe7-5f84-4324-9d93-ad98933c9b3a)
- Identifier: [pstsc_09844_e15ca556987b120ed80a50444194464a](https://cho-qa.libraries.psu.edu/?utf8=%E2%9C%93&search_field=title&q=pstsc_09844_e15ca556987b120ed80a50444194464a)
- Identifier: [pstsc_09844_e15ca556987b120ed80a50444194464a_00003_001](https://cho-qa.libraries.psu.edu/?utf8=%E2%9C%93&search_field=title&q=pstsc_09844_e15ca556987b120ed80a50444194464a_00003_001)
- Filename: [pstsc_09844_e15ca556987b120ed80a50444194464a_00003_001_text.txt](https://cho-qa.libraries.psu.edu/?utf8=%E2%9C%93&search_field=title&q=pstsc_09844_e15ca556987b120ed80a50444194464a_00003_001_text.txt)

awead:
Possibly related to #833

awead:
To reproduce:
Search for filename or an identifier and it should works.

awead:
Searching the complete filename works, however, partial names like pstsc_06352_b01f01 do not. Verify that Solr config change is made in QA. You should not see the ICU tokenizer for tesim fields.

awead:
After the correct solr upgrade this is working, but the search result ranking might be misleading.

PR #906, opened 2019-03-22
Updating to fix searching

cam156:
## Description

There were a number of issues with how the catalog was being searched.

Connected to #839
ref https://github.com/psu-stewardship/chef/pull/856

## Changes

* Updated the catalog controller to allow for search fields and all fields
* updating the tokenizer in the solr schema to break on all punctuation and underscore

### DIFF
--- a/solr/config/schema.xml
+++ b/solr/config/schema.xml
@@ -151,7 +151,7 @@
 		</fieldtype>
 		<fieldType name="text_en" class="solr.TextField" positionIncrementGap="100">
 			<analyzer>
-				<tokenizer class="solr.ICUTokenizerFactory"/>
+				<tokenizer class="solr.PatternTokenizerFactory" pattern="[\W_]"/>
 				<filter class="solr.ICUFoldingFilterFactory"/>
 				<filter class="solr.EnglishPossessiveFilterFactory"/>
 				<filter class="solr.EnglishMinimalStemFilterFactory"/>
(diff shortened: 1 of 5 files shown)


solr/config/schema.xml
```diff
@@ -151,7 +151,7 @@
 		</fieldtype>
 		<fieldType name="text_en" class="solr.TextField" positionIncrementGap="100">
 			<analyzer>
-				<tokenizer class="solr.ICUTokenizerFactory"/>
+				<tokenizer class="solr.PatternTokenizerFactory" pattern="[\W_]"/>
```
awead: @cam156 if we're tokenizing on whole words, then will this break partial-word searching? We solved the underscore problem in SS with a mapping file that replaced the underscores with spaces in a search field. https://github.com/psu-stewardship/scholarsphere/commit/45babcedf950b40048d687cc5fcd78801664e83d I'm not sure I like that solution here though.

cam156: @awead that solution does not allow you to search for _ and space aka hello_world or hello world. I'll look at keeping both tokenizers like we discussed in standup.

awead: @cam156 Sounds good. I'm ok with the PR as-is. If you can add the additional tokenizer without any ill effects, then that's good. Otherwise, put in another ticket to track this issue for future enhancements.

Issue #824, opened 2019-02-22
Duplicate Identifiers Bug


ntallman:
I accidentally made a duplicate collection for the work, not realizing old data would remain when a new branch was deployed. I assigned the same identifier (alternate_id) for a new collection and CHO didn't complain. The identfier (alternate_id) field maps to the valkyrie property that is supposed to be unique

- https://cho-qa.libraries.psu.edu/catalog/763b12bd-2add-4321-a0b8-1afe984cafff
- https://cho-qa.libraries.psu.edu/catalog/2bff54a4-03e3-4e1a-a171-2d8cc079d6cc

ntallman:
If this is a bug, maybe it shouldn't be? Thinking about the valkyrie property. If you had to delete a work and then later add the same work, it would have the same local identifier. If Valk thought it needed to be universally unique, wouldn't that mean it should block you using the same identifier? We would have to use the same local identifier. This may be symptomatic of our approach to multiple managing identifiiers.

awead:
TODO:
* investigate for possible Valkyrie bug, and report as needed
* if this is a bug in Valkyrie, fix it in the Valkyrie gem
* if it is not a bug in Valkyrie fix it locally
* verify that a user cannot create a collection with the same alternate identifier

awead:
Valkyrie _assumes_ that your alt. ids are unique, so it's up to us to do the enforcement.

PR #907, opened 2019-03-22
Adds unique validation to alternative identifiers

awead:
## Description

Valkyrie does not enforce any restrictions about alternate identifiers. It assumes that the downstream application will do whatever work is needed. To that end, we need to ensure that the alternate identifiers we provide are unique system-wide.

Connected to #824 

## Changes

* Creates a `Unique` validator and assigns it to the `alternate_ids` property in the data dictionary
* Changes the way validators are invoked from the change set so that the current change set is passed to the validate, as well as the field it is checking. This enables us to determine if the field is being updated from an existing value, or is a new, unsaved record.


spec/cho/validation/member_spec.rb
```diff
@@ -3,19 +3,35 @@
...
     end
 
     context 'with a non-existent id' do
       subject { described_class.new('non-existent-id') }
 
-      its(:exists?) { is_expected.to be(false) }
+      its(:exists_by_id?) { is_expected.to be(false) }
+    end
+  end
+
```
cam156: The validation of the exists? alias is untested Are you intending to keep that alias?

spec/cho/work/submissions/edit_spec.rb
```diff
@@ -3,7 +3,7 @@
 require 'rails_helper'
 
 RSpec.describe 'Editing works', type: :feature do
-  let(:resource) { create(:work, :with_creator, title: 'Work to edit', work_type_id: work_type.id) }
+  let(:resource) { create(:work, :with_creator, :with_metadata, title: 'Work to edit', work_type_id: work_type.id) }
```
cam156: This seems like an odd change.  The setup changed, but nothing in the test changed.  It looks to me like you are now requiring the alternate id, but it is not required to save.

awead: @cam156 the alt. id was empty previously. I needed to add it to verify we can update works with one. The `:with_metadata` trait is really a nudge that we ought to start filling in all the fields in our forms to get an idea of things will break or not. The update vs. create issue with unique alt. ids never would have surfaced if I hadn't put one on a resource.

spec/cho/validation/member_spec.rb
awead: I've added a test.

PR #911, opened 2019-03-22
Adding a new reindex task

cam156:
Utilizes the migration example code form valkyrie to reindex in solr from postgres

### DIFF
--- a/lib/tasks/index.rake
+++ b/lib/tasks/index.rake
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+namespace :cho do
+  namespace :index do
+    desc 'Batch load works from a directory of csv files'
+    task reindex_all: :environment do
+      postgres_adapter = Valkyrie::MetadataAdapter.find(:postgres)
+      solr_adapter = Valkyrie::MetadataAdapter.find(:index_solr)
+      postgres_query_serice = postgres_adapter.query_service
+      solr_persister = solr_adapter.persister
+      index_all(query_service: postgres_query_serice, persister: solr_persister, model: Collection::Archival)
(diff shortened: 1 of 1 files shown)


lib/tasks/index.rake
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+namespace :cho do
+  namespace :index do
+    desc 'Batch load works from a directory of csv files'
+    task reindex_all: :environment do
+      postgres_adapter = Valkyrie::MetadataAdapter.find(:postgres)
+      solr_adapter = Valkyrie::MetadataAdapter.find(:index_solr)
+      postgres_adapter.query_service.find_all.each { |resource| solr_adapter.persister.save(resource: resource) }
```
awead: @cam156 do we need to do a buffered persister so we aren't saving to solr each time?

cam156: @awead we could do a buffered persister.  I was just throwing something simple up.  Maybe we should create a ticket that has our desired reindexing scheme.  I tried running the code on the qa server and it did not complete because of the connections between the records.  I created an update, bu that does not run because of the agent nil role error.

cam156: It might be better to do a save_all like figgy is doing for a slice: https://github.com/pulibrary/figgy/blob/c17fa9e90cf4b3260bb349e78f9bdaab6388007a/app/services/reindexer.rb

awead: @cam156 sounds good. Let's get something working and ticket/plan for an improvement.

cam156:
@awead this should be in a working state.  It is getting way more complicated than I was expecting now with having to buffer data into solr.

Issue #721, opened 2018-11-13
"Index Type" in Data Dictionary does not work for Collections


ntallman:
Due to display transformations, editing the "Index Type" (no_facet/facet) for the `member_of_collection_ids` field has no effect.

Related to #459
Related to #689

awead:
TODO:
* make the index_type field of `member_of_collection_ids` control the appearance of the facet for Collections

PR #914, opened 2019-03-25
Configuring member_of_collection_ids in the data dictionary

cam156:
## Description

Removing duplicate member of collection fields and allowing the data dictionary one to display correctly.

Connected to #721

## Changes

* Configuring member_of_collection_ids in the data dictionary
* Updates the data dictionaries to have collections transformation
* Adds render_link_to_collection as a transformation
* Updated render_link_to_collection to display both multivalue and single value for facet
* change field displayed for member of collection from member_of_collection_id_ssim to member_of_collection_tesim since that is a default for a field


app/blacklight/local_helper_behavior.rb
```diff
@@ -1,10 +1,17 @@
...
-    raw(options[:document].member_of_collections.map do |document|
-      link_to_document(document, document_show_link_field(document))
-    end.join(', '))
+  def render_link_to_collection(options_or_value = {})
+    if options_or_value.is_a? Hash
+      raw(options_or_value[:document].member_of_collections.map do |document|
+        link_to_document(document, document_show_link_field(document))
+      end.join(', '))
+    else
+      doc = SolrDocument.find(options_or_value.sub(/^id-/, ''))
```
awead: @cam156 shouldn't we already have the solr document, i.e. `@document`, in this context?

cam156: @awead this is the collection's solr document.  The information is coming back in the facet field and is only the id.

Issue #572, opened 2018-06-25
Acknowledgements on Collection Landing Page


ntallman:
As a metadata librarian/cataloger/collections librarian, when describing a collection;
I would like to record and name the contributions of the people doing the curation, digitization, description, migration, and other work which has led to the collection becoming available in CHO;
so that all may be recognized for their contributions to the collection.

Ruth K-T.

Child of #483

awead:
TODO:
* create "acknowledgements" property in the data dictionary, as a optional text field
* assign acknowledgements to the collection schema
* present field in the collection show page (a.k.a. the landing page) above the listing of items in the collection

PR #918, opened 2019-03-27
Paragraph-heading transformation for acknowledgments

awead:
## Description

Creates a new kind of transformation that displays a property as a separate paragraph with a heading. This is used in conjunction with a new acknowledgments property that is available to collections.

Connected to #572 

## Changes

* add heading transformation
* acknowledgments property to collections
* MetadataFieldPresenter for displaying contents of field as separate paragraphs

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -46,3 +46,4 @@ Metrics/BlockLength:
 Rails/Presence:
   Exclude:
     - 'app/cho/indexing/creator.rb'
+    - 'app/blacklight/metadata_field_presenter.rb'
--- a/app/blacklight/catalog_controller.rb
+++ b/app/blacklight/catalog_controller.rb
@@ -57,8 +57,12 @@ class CatalogController < ApplicationController
       catalog_field = map_field.solr_search_field
       catalog_label = map_field.display_name || map_field.label.titleize
       catalog_helper = map_field.display_transformation == 'no_transformation' ? nil : map_field.display_transformation.to_sym
-      config.add_index_field catalog_field, label: catalog_label, helper_method: catalog_helper
-      config.add_show_field catalog_field, label: catalog_label, helper_method: catalog_helper
+
+      unless map_field.display_transformation == 'paragraph_heading'
+        config.add_index_field catalog_field, label: catalog_label, helper_method: catalog_helper
+        config.add_show_field catalog_field, label: catalog_label, helper_method: catalog_helper
+      end
+
       config.add_search_field(map_field.label) do |field|
         # solr_parameters hash are sent to Solr as ordinary url query params.
         field.solr_parameters = { 'spellcheck.dictionary': field.label,
--- a/app/blacklight/local_helper_behavior.rb
+++ b/app/blacklight/local_helper_behavior.rb
@@ -15,4 +15,9 @@ module LocalHelperBehavior
       doc['title_tesim'].first
     end
   end
+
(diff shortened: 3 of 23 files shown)


awead:
@rschenk These are excellent points. I'll address each one:

1. `DisplayTransformation` was initially a factory pattern that was built along with `ControlledVocabulary` and `Validation`. They were build at a time when we didn't know _exactly_ what we wanted to implement yet, we just knew we were going to have a bunch of these types of objects and the factory pattern seemed to make the most sense. The validation and controlled vocabs. matured in the last year and the pattern largely works; although, we're completely open to revisiting it. The display transformations have not matured because we really haven't had anything that needed transformation, until now...
2. This brings us to Blacklight, whose default implementation is to (basically) throw everything in a helper. Because helpers are global, the pattern is to create your own local helper module and define the method in there that either overrides the existing Blacklight one, or does whatever custom behavior you need to render your field. This works for cases where you're sticking to the definition list display. If you need to do something "fancier," then you have to look to patterns outside Blacklight's helper-based transformations. This brings me to:
3. We use presenters to wrap up more complicated display logic and keep views free of excess Ruby code. Blacklight has presenters too, but I've only been able to leverage them for changes to the page's display, and not individual fields. I could be wrong, though, and maybe there's something I'm missing, but that led me to create a simple PORO for displaying a metadata field where I could put the logic I needed to show a paragraph with a heading. It's supposed to do instancy-type stuff. Maybe I didn't do it right? It uses the existing solr document and a field, pulls out the values and displays them.
4. The confusion here is due to the fact that Blacklight has helper-based "transformations" but we have others that don't quite fit into a helper-oriented pattern. `render_link_to_collection` is one of these helpers, while the `paragraph_heading` is not and is a pattern unto its own.

I like were you are going with your last suggestions, but maybe not going entirely away from the helper pattern because it's so baked into Blacklight. Definitely having a presenter/partial per non-helper transformation make a lot of sense.

It also might we worth digging into Blacklight more, or asking the community, I'm sure we're not the first people to run into this scenario.

rschenk:
@awead I think what rubs me the wrong way about `MetadataFieldPresenter` is that eventually we're going to have multiple different types of fields, all of which might want to be presented in a different way. When I said it doesn't do any "instancy" stuff what I was getting at was, the way that it's a single class right now, I can foresee this class becoming littered with `case field.display_transformation`. 

I think what threw me off was, I had assumed that was the purpose of `DisplayTransformation`, but I guess the intention of the `DisplayTransformation` classes is to transform stuff before it gets saved into our storage?

I could imagine something like this...
```
class MetadataFieldPresenter::Base
  def initialize(field, document)

  delegate blank, present, etc etc
 
  def label

  def content

  def to_partial_path
end

class MetadataFieldPresenter::ParagraphWithHeading < MFP::Base
  def paragraphs
end

# in the future...

# (this would be #899 and we'd name it better)
class MetadataFieldPresenter::ParagraphWithoutHeading < MFP::ParagraphWithHeading
end

class MetadataFieldPresenter::DOI < MFP::Base
  def content
    "https://doi.org/#{super}"
   end
end
```

Thoughts?

awead:
@rschenk Yes, I see what you mean. Maybe the field presenter is, in fact, the display transformer and the factory should building instances of a presenter/transformer class. The role of the `DisplayTransformation` class is to do exactly that: change up things for display. The simplest example might be take a state code like `PA` and transform it to `Pennsylvania`.

Issue #482, opened 2018-04-16
Work Landing Page Enhancements


ntallman:
As a remote scholar using primary source materials in CHO,
I would like to easily tell how to download a single file for the entire work on the page I am viewing,
So that I can quickly download what I need and begin my research.

In addition, it would be helpful to see a sample image from the work and know how I can download one or more high-resultion images of individual pages, should I need them.

Also, instead of Files for the heading to the list of file sets, please use Parts.

ntallman:
I think @mtribone may have already added thumbnails to works? (Or was that just file sets?)

Re: the downloads and files, was thinking of a download button for the representative file for the work, if present. Works could have hundreds of file sets, unordered lists could get unruly. We might want to link out to a new page where a table of file sets. The table could have the identifier, title, date, and thumbnail. It could also have a link to the fileset show page. (Bonus: each row could have a checkbox for bulk download of file sets.)

awead:
TODO:
* change "Files" heading to "Parts"
* suppress representative file set in work listing
* button to download the representative access file ("Download work")
* button to download text file ("Download plain text")
* thumbnails are already visible

PR #919, opened 2019-03-27
Buttons download representative work & plain text

rschenk:
## Description

On the Work landing page, add dedicated buttons to download the representative access file and the full plain text. If the work has a thumbnail, these buttons are added in the left column just below the thumbnail; if the work has no thumbnail, there is no left column so the buttons are added above the work's title. The styles need a little love... it would be great if we could do a little flexbox magic and lay the buttons out vertically if there's a thumbnail, and horizontally if not. @mtribone, perhaps you and I could pair on that while @awead is out tomorrow (Thursday).

_Please note,_ this functionality works through a new field in the solr document, so you will need to rebuild your solr index to see it.

Connected to #482 

## Changes

* Index representative file set id in solr
* Add new partial for the download buttons.
* Hack in some new factory traits to make this testable... Adam and I would like to refactor the factories to make this cleaner at some point

### DIFF
--- a/app/views/catalog/show.html.erb
+++ b/app/views/catalog/show.html.erb
@@ -13,14 +13,12 @@
     <% if presenter(@document).thumbnail.exists? %>
       <div class="col-xs-12 col-md-4">
         <%= render 'thumbnail', document: @document, document_counter: 0 %>
-        <%= render 'download_representatives', document: @document %>
       </div>
       <div class="col-xs-12 col-md-8">
         <%= render_document_main_content_partial %>
       </div>
     <% else %>
       <div class="col-xs-12">
-        <%= render 'download_representatives', document: @document %>
         <%= render_document_main_content_partial %>
       </div>
     <% end %>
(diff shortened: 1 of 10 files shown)


app/views/catalog/show.html.erb
```diff
@@ -13,12 +13,14 @@
     <% if presenter(@document).thumbnail.exists? %>
       <div class="col-xs-12 col-md-4">
         <%= render 'thumbnail', document: @document, document_counter: 0 %>
+        <%= render 'download_representatives', document: @document %>
```
awead: @rschenk what do you think about `<%= render 'download_representatives', document: @document if document.has_representative? %>` and removing the `if` block in the `download_representatives` partial?

rschenk: Sounds fine to me. I was following the pattern in the  `<%= render 'thumbnail' ` line above, which also has an if as its first line in the partial. I don't have a strong feeling either way

Issue #739, opened 2018-12-03
MVP Home Page


awead:
~~CHO needs a complete homepage with appropriate links, text, and other content to be determined after engaging with stakeholders. It is assumed that some of the content will be dynamic, depending on existing content in CHO, as well as static, and possible user-created, such as editable sections of the page, similar to what we have in Scholarsphere.~~

* CHO needs an MVP homepage that more barebones, but still elegant. A minimalistic design that emphasizes discovery. Content does not *need* to be dynamically generated for MVP. Editable blocks are not needed for MVP, at most 1 block of introductory text would be needed. 
* Merging related ticket (#714): Update the home page view to include an h1 for accessibility and code quality.

Inspiration
- [NYPL](https://digitalcollections.nypl.org/)
- [Duke](https://repository.duke.edu/dc?)
- [Oregon Digital](https://oregondigital.org/catalog/)
- [UCSD](https://library.ucsd.edu/dc/)

Related: #768

[Updated by NT]

ntallman:
Merging related ticket (#714): Update the home page view to include an h1 for accessibility and code quality.

mtribone:
![cho-home-2nav](https://user-images.githubusercontent.com/4163828/54930418-9ecd9680-4eed-11e9-8add-a7b73c397a33.png)

mtribone:
![cho-home-2nav-min](https://user-images.githubusercontent.com/4163828/54930437-a856fe80-4eed-11e9-9024-4a9d55ed9d76.png)

mtribone:
![cho-home-kee-ullered](https://user-images.githubusercontent.com/4163828/54948417-f8948780-4f12-11e9-9cac-468669a45214.png)

Issue #810, opened 2019-01-31
MVP Navigation Menu based on Role


ntallman:
Before MVP is released, we need a separate nav menu for public and psu users (non-application users.) The current navigation menu should be seen by application users.

Non-application users (public and logged in PSU users) should the following items on a nav bar:
- Home
- About
- Browse Collections (faceted search for collections)
- Browse Works (faceted search for works)
- Bookmarks
- History
- Log Out (if applicable)

Is there a good way to have these menu items also appear for application users?

#788 is a future enhancement to this.

PR #932, opened 2019-04-01
New Header and Navigation Menu for Home and Secondary Pages

mtribone:
Adds another partial for splitting up the menu for profile, primary, and secondary navigation. Updates the catalog controller to add 2 non-working links to the new mwnu for My Works and My Collections and adds partials for both. Includes the layout behavior for the controllers so that they pick up the override for the container classes from CHO and not Blacklight

Adds additional SCSS file for the header styles and a variable for the PSUL blue. Breaks the header/banner into two parts for banner, profile nav, and primary navigation. Updates the layout of the header for Bootstrap columns from mobile to desktop.

## Description

Customizes CHO header and navigation to make it visually more like Libraries and less like default Blacklight

Connected to #739 

## Changes

* Adds new graphics for header background and Libraries logo
* Splits up current menu into 2 sections. One menu for profile related item and another menu for the creation of works and collections.
* Semantic updates to cho layout partial. Replaces `<div>` with `<main>`, `role='banner'` for the div containing the branding
* New color variables for Libraries branding
* Adds nav action in Blacklight catalog controller for upcoming "My Works" and "My Collections" navigation items
* Overrides additional Blacklight partials to use our container classes for fluid layout
* Adds and removes Bootstrap classes where needed


mtribone:
Screen Shot 2019-04-01 at 3 31 50 PM

mtribone:
Screen Shot 2019-04-01 at 3 32 28 PM

mtribone:
Screen Shot 2019-04-01 at 3 32 39 PM

app/cho/agent/resources_controller.rb
```diff
@@ -4,6 +4,8 @@ module Agent
   class ResourcesController < ApplicationController
     include ValkyrieControllerBehaviors
 
+    helper LocalHelperBehavior, LayoutHelperBehavior
+
```
awead: @mtribone we seem to need these everywhere. Would it make more sense to put them in `ApplicationController`? That way all other controllers would inherit from it. The only counterpoint would if we wanted to have a controller that _shouldn't_ inherit these helpers.

app/views/layouts/cho.html.erb
```diff
@@ -19,6 +19,7 @@
     <title><%= render_page_title %></title>
     <%= opensearch_description_tag application_name, opensearch_catalog_url(format: 'xml') %>
     <%= favicon_link_tag %>
+    <%= stylesheet_link_tag 'https://use.typekit.net/uiy7uef.css' %>
```
awead: @mtribone is this a stable, versioned resource? What would happen if it went "away" or changed? It's also totally acceptable to punt on this right now.

app/assets/stylesheets/cho.scss
```diff
@@ -1,3 +1,3 @@
 @charset "UTF-8";
 
-@import 'cho/variables', 'cho/blacklight', 'cho/data_dictionary_fields', 'cho/facets', 'cho/footer', 'cho/form', 'cho/scaffolds';
+@import 'cho/variables', 'cho/blacklight', 'cho/data_dictionary_fields', 'cho/facets', 'cho/footer', 'cho/form', 'cho/header', 'cho/scaffolds';
```
awead: @mtribone how would you feel about new lines for each `@import`? It's not pretty, but it'll make future diffs easier to read. On the other hand, if this isn't going to change much or will go away, feel free to leave it as-is.

app/cho/agent/resources_controller.rb
mtribone: @awead Yeah it seemed like I needed to add one right after another when I was moving through the site.

app/views/layouts/cho.html.erb
mtribone: This is the typeface for the header that the University uses as its 'official' typeface. We get it 'free' as part of the University's relationship with Adobe. It is possible that the contract might end in the future, but the CSS does have a fallback.

app/blacklight/catalog_controller.rb
```diff
@@ -185,5 +185,8 @@ class CatalogController < ApplicationController
     # Configuration for autocomplete suggestor
     # config.autocomplete_enabled = true
     # config.autocomplete_path = 'suggest'
+
+    config.add_nav_action(:my_works, partial: 'blacklight/nav/my_works')
```
mtribone: @awead These are here in prep for #810 but are non-functioning links.

app/views/layouts/cho.html.erb
awead: @mtribone ok, that's fine for now then. My concern was we might want to import that css directly into the app so we don't have to rely on the external source, but it sounds like we want that external link so it's from a common location.

mtribone: @awead oh. I gotcha. let me see if that works.

Issue #899, opened 2019-03-19
Textual content from curator


awead:
Add additional field to data dictionary as a text type for multiple paragraphs. Field should only be available to collections.

Child of #483

awead:
TODO:
* add `narrative` field to the collection schema
* render the same as the acknowledgment field  (see #572)
* format in plain text

PR #934, opened 2019-04-02
Paragraph transformation for curator’s textual narrative

rschenk:
## Description

Creates a new kind of transformation that displays a property as one or more paragraphs. This is used with the new `narrative` property on collections that allows curators to provide text content.

Connected to #899 

## Changes

* Add new `narrative` field to data dictionary (please double check I got all the options right)
* Add new display transformation partial for a paragraph type
* Narrative gets displayed above Acknowledgments (and metadata list) on Collection pages... displaying it below the Acknowledgments made it look like it was a part of that section

### DIFF
--- a/app/blacklight/catalog_controller.rb
+++ b/app/blacklight/catalog_controller.rb
@@ -53,12 +53,17 @@ class CatalogController < ApplicationController
     # @todo configuration of linked field indexing
     config.add_facet_field 'creator_role_ssim', label: I18n.t('cho.field_label.role')
 
+    fields_excluded_from_metadata_list = %w[
+      acknowledgments
+      narrative
+    ]
+
     DataDictionary::Field.all.sort_by(&:created_at).each do |map_field|
       catalog_field = map_field.solr_search_field
       catalog_label = map_field.display_name || map_field.label.titleize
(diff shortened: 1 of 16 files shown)


app/blacklight/catalog_controller.rb
```diff
@@ -53,12 +53,17 @@ class CatalogController < ApplicationController
     # @todo configuration of linked field indexing
     config.add_facet_field 'creator_role_ssim', label: I18n.t('cho.field_label.role')
 
+    fields_excluded_from_metadata_list = %w[
+      acknowledgments
+      narrative
+    ]
+
```
rschenk: I waffled back and forth between excluding these based on `display_transformation` type, as was done before, or by the field name. Although arguably less DRY, doing it by the field name felt a little easier to understand for some reason. Thoughts @awead?

awead: This is fine for now. I'll have some more comments on the PR in general, but I think there ought to be a way off-load this decision to something else. The CatalogController shouldn't have to know these things (ideally).

awead:
@rschenk This works, so I'm inclined to merge it, but with a subsequent refactoring, if you're willing. I think the mail issue is around CatalogController having to know too much. The display transformation should be doing some more lifting.

rschenk:
@awead sounds good, and agreed

PR #937, opened 2019-04-05
Making tests work if the label for title changes

cam156:


spec/views/archival_collections/edit.html.erb_spec.rb
```diff
@@ -20,7 +20,7 @@
       assert_select 'input[name=?]', 'archival_collection[visibility]'
       # Added to make sure accessibility changes are in place
       assert_select 'legend', 'Basic Metadata'
-      assert_select 'label', /Title\s.* required/
+      assert_select 'label', for: 'library_collection_title', text: /\s.* required/
```
awead: @cam156 This needs to be `assert_select 'label[for=?]', 'archival_collection_title', text: /\s.* required/`. The `for:` I don't think is doing anything here.

cam156: Good catch! It seems the test is not testing what I expected it to.  I'll fix it up.

Issue #885, opened 2019-03-14
FindUsing does not escape punctuation


awead:
Custom sql queries in FindUsing do not properly escape punctuation. When searching for an agent's name that has an apostrophe, there's an error:

```
     ActiveRecord::StatementInvalid:
       PG::SyntaxError: ERROR:  syntax error at or near "Hara"
       LINE 1: ...":["Gabrielle"]}' AND metadata @> '{"surname":["O'Hara"]}' A...
                                                                    ^
       : SELECT * FROM orm_resources WHERE metadata @> '{"given_name":["Gabrielle"]}' AND metadata @> '{"surname":["O'Hara"]}' AND internal_resource = 'Agent::Resource';
     # ./app/valkyrie/find_using.rb:23:in `run_query'
     # ./app/valkyrie/find_using.rb:19:in `find_using'
```

PR #940, opened 2019-04-08
Escape characters when querying Postgres

awead:
## Description

Fixes bug when querying Postgres with values that have punctuation.

Connected to #885

### DIFF
--- a/spec/valkyrie/find_using_spec.rb
+++ b/spec/valkyrie/find_using_spec.rb
@@ -71,4 +71,17 @@ RSpec.describe FindUsing do
       expect(results.first.id).to eq(persisted_sample.id)
     end
   end
+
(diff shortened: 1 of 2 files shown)


spec/valkyrie/find_using_spec.rb
```diff
@@ -71,4 +71,13 @@ class SimilarResource < Valkyrie::Resource
       expect(results.first.id).to eq(persisted_sample.id)
     end
   end
+
+  context 'when escaping punctuation in the search term' do
+    subject { saved_resource }
```
cam156: Shouldn't the subject be :retrieved_resource?

awead: @cam156 yes, that doesn't look right. I've re-done the test.

Issue #797, opened 2019-01-31
Implement Repeating Fields on Metadata Forms


ntallman:
Data dictionary allows fields to be defined as multi-valued. CSV ingest has the ability to accept multiple values/value-pairs, but our metadata edit forms do not have the ability to add repeating fields. #98 says we used an older rails method to do this until we were out of MVP, but if it was implemented, it doesn't seem to be working anymore.

PR #943, opened 2019-04-08
Installing eslint

awead:
## Description

Uses the standard settings for linting JS files.

Connected to #797

## Note

I'm punting on Travis integration because we'll likely be moving to CircleCI and that has better options for building Ruby+Node apps.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,17 @@
+{
+  "env": {
+    "browser": true,
+    "es6": true
+  },
+  "extends": "standard",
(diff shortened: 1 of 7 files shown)


awead:
@cam156 you need to run `yarn lint`, so I should probably put something in the wiki docs for that. Later, it'll get run during the CI process, probably with CircleCI.

awead:
@cam156 I've added a blurb here https://github.com/psu-libraries/cho/wiki/Development-Setup#linters

Issue #802, opened 2019-01-31
Ordering of Works


ntallman:
Works, file sets, and files are not sorted properly on display. One approach would be to sort on the identifier, regardless of what is displayed. That might be easier for MVP. Or, preserve the order from the CSV? (Both, depending on ingest method?) Implementing [ORE ordering](https://github.com/duraspace/pcdm/wiki/PCDM-2.0#ordering) might be the 1.x solution?

awead:
TODO:
* sort works on identifier
* verify works are displayed in a collection in the order according to their identifier (alternative_ids field)

PR #944, opened 2019-04-09
Order collection’s works by their position in CSV

rschenk:
## Description
Works are explicitly ordered by their position in the CSV. If multiple CSVs contain works referencing the same collection, these works are explicitly ordered by the order in which those CSVs were ingested, i.e. works ingested at a later time will be explicitly ordered as such.

Connected to #802 

## Changes

* Add order_index to Work::Submission and SubmissionChangeSet
* CsvDryRun uses the index of the work and the number of existing members in that works' home collection to generate said order_index


spec/cho/work/import/csv_dry_run_spec.rb
```diff
@@ -178,6 +179,62 @@
...
+          )
+        end
+
+        before do
+          # Create two existing works in that collection
+          create :work, home_collection_id: collection.id
+          create :work, home_collection_id: collection.id
+        end
+
+        it 'increments the order_index starting at 1' do
```
awead: @rschenk it starts at 2, right?

rschenk: Good catch!

Issue #895, opened 2019-03-19
Add Blacklight gallery


awead:
Displays search results in a "Gallery" format

Child of #483

rschenk:
The most recent release of the blacklight-gallery gem does not support Blacklight 7.

The current HEAD of the master branch of the blacklight-gallery repo _does_ have support for blacklight 7, but there is no release tag. We could install the gem straight off the repo's master but it seems like it may be more prudent to punt on this until it's officially released?

awead:
I'll enquire about the status of the release. Let's hold off until we know.

awead:
From code4lib#blacklight: "We aren't using BL 7 for our project that uses blacklight-gallery yet.
So, we did some work to show that BL 7 would work in theory, but I'm not sure we've done extensive testing."

Given that, I think we can go ahead and use master, provided it doesn't completely blow anything up, and see how it works out. If we find bugs, we can ticket them either here or in the bl-gallery repo.

PR #961, opened 2019-04-11
Add Blacklight gallery

rschenk:
## Description

Adds the blacklight-gallery gem, but there are several things of note here

The current official release of blacklight-gallery gem does not support Blacklight 7, therefore I had to install it from the master branch of its GitHub repository. We are using and alpha version of the gem.

The masonry layout looks quite terrible and our image thumbnails are not properly cropped/resized to the columns as would be expected. I personally am hesitant to deploy this, so I will leave it up to the rest of the team to decide whether it is ready go now, or needs some UI attention before it is deployed.

On a technical note, CHO's nonstandard app folder layout caused the installation generators for both blacklight-gallery and its dependency opeanseadragon to crash. I read through their code, and manually inserted the installation code into its respective locations in CHO. 

Connected to #895 

## Changes

* install alpha version of blacklight-gallery directly from github repo
* manually install the blacklight-gallery and openseadragon gems because install generators crashed
* inserted blacklight-gallery and openseadragon css's into the asset pipeline — @mtribone could you please confirm that I did that in a pleasing way

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,6 +19,7 @@ gem 'twitter-typeahead-rails', '0.11.1.pre.corejavascript'
 
 # Supporting gems
 gem 'bagit'
+gem 'blacklight-gallery', git: 'https://github.com/projectblacklight/blacklight-gallery'
 gem 'bootsnap', require: false
 gem 'cancancan'
 gem 'coffee-rails', '~> 4.2'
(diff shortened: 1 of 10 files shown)


mtribone:
@rschenk I think you got it right with the asset pipeline.

awead:
Although this has some UI display problems, will address these in later releases.

Issue #974, opened 2019-05-09
Implementing authenticated visibility


awead:
When the user selects "PSU Access" as a visibility, then the resource should be restricted to only users who are authenticated to Penn State.

TODO:
If we follow the Hyrax paradigm, there would be a "authenticated" group that would be used to control this on the resource.

PR #986, opened 2019-05-21
Enforce authenticated access

awead:
## Description

Defines access levels, previously known as _visibility_, on collections, and creates a central vocabulary and transaction for assigning them. Using Blacklight's access controls gem, these levels are used to restrict access to collections, works, and file sets based on whether or not a user is authenticated to Penn State or not.

Connected to #974 

## Changes

* rename visibility to access level
* create a RDF and repository vocabularies
* insert new operation to apply access levels to collections
* removes dependency on Valkyrie's access controls
* installs Blacklight's access controls
* enforces search, view, and edit restrictions on resources


app/cho/transaction/operations/shared/apply_access_level.rb
```diff
@@ -0,0 +1,22 @@
...
+  module Operations
+    module Shared
+      class ApplyAccessLevel
+        include Dry::Transaction::Operation
+
+        def call(change_set)
+          return Success(change_set) if change_set.try(:access_level).blank?
+          return Failure(change_set) unless change_set.valid?
+
+          change_set.resource.read_groups += [change_set.access_level]
```
rschenk: Is there a need to call `uniq` on this list? And/or is it a Set already? And/or does it matter if there are duplicates in this list?

app/cho/transaction/operations/shared/apply_access_level.rb
```diff
@@ -0,0 +1,22 @@
...
+          Success(change_set)
+        rescue StandardError => e
+          change_set.errors.add(:apply_access_level, e.message)
+          Failure(change_set)
```
rschenk: Helpful to use `Rejection` here?

app/devise/user.rb
```diff
@@ -2,13 +2,14 @@
...
   include Blacklight::User
+  include Blacklight::AccessControls::User
   devise :http_header_authenticatable
 
   def groups
     return group_list.split(';?;') unless groups_last_update.blank? || ((Time.now - groups_last_update) > 24 * 60 * 60)
 
     populate_attributes
-    group_list.split(';?;')
+    group_list.split(';?;') + default_groups
```
rschenk: I'm not quite sure I follow what's going on here, but I just want to make sure that the `+ default_groups` bit does not also need to be in the guard clause up in line-9

app/cho/transaction/operations/shared/apply_access_level.rb
awead: Great catch! In short, no. A `Set` already enforces uniqness, and it probably wouldn't matter anyway if there were dups, but it's nice to know that there won't be.

app/cho/transaction/operations/shared/apply_access_level.rb
awead: Yep, I think so. I'll change that...

app/devise/user.rb
awead: `group_list` is a method from the PsuDir gem, and is completely *not* obvious at all here. It's a listing of groups from our LDAP server. We split those and add the default groups to that. I think some comment might be helpful here.

PR #990, opened 2019-05-22
Enforce access to downloads

awead:
## Description

Adds ability checks to the downloads controller and fixes a bug where administrators were not able to download unredacted preservation files.

Connected to #974 

## Changes

* access abilities added to DownloadsController
* administrators can download unredacted preservation files
* file set factory now creates files with appropriate suffixes
  with multiple files supported


awead:
File set permissions are not getting updated when the work or collection's access rights change.

awead:
Above will be fixed in #1026

Issue #837, opened 2019-02-28
Confirmation when deleting agents


ntallman:
When deleting an agent (individual or batch),
I should see confirmation messages as when deleting works and collections,
so I don't accidentially delete an agent.

I should not be able to delete an agent if a work or collection has that agent listed in their metadata.

PR #991, opened 2019-05-22
Can delete Agents only if they have no members

rschenk:
## Description

If an Agent has any members (for example, works they're listed as a creator upon), the delete link is hidden. Otherwise the delete link is visible, with a javascript "are you sure...?" confirmation.

Connected to #837 

## Changes

* Add method to fetch member ids of an Agent
* Update Ability class to ensure an Agent has no members before allowing it to be deleted

### DIFF
--- a/app/devise/ability.rb
+++ b/app/devise/ability.rb
@@ -10,6 +10,7 @@ class Ability
     base_permissions
     authenticated_permissions
     admin_permissions
+    cannot_delete_agent_with_members
   ]
 
   def base_permissions
@@ -27,5 +28,11 @@ class Ability
 
   def admin_permissions
     can :manage, :all if current_user&.admin?
+  end
+
+  def cannot_delete_agent_with_members
+    cannot :delete, Agent::Resource do |agent|
+      agent.member_ids.any?
+    end
   end
 end
--- a/spec/cho/agent/agent_spec.rb
+++ b/spec/cho/agent/agent_spec.rb
@@ -11,6 +11,8 @@ RSpec.describe Agent::Resource, type: :model do
                       given_name: 'sally',
                       surname: 'brown' }
 
+  let(:saved_model) { Valkyrie.config.metadata_adapter.persister.save(resource: model) }
+
   it_behaves_like 'a Valkyrie::Resource'
 
   it { is_expected.to respond_to(:given_name) }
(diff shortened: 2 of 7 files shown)


spec/cho/agent/agent_spec.rb
```diff
@@ -30,4 +31,23 @@
...
+      end
+    end
+
+    context 'when the agent has members' do
+      let!(:work) { create :work_submission, :with_creator }
+
+      it "returns an array of its members' Valkyrie IDs" do
+        agent = Agent::Resource.find(work.creator.first.fetch(:agent))
+
+        expect(agent).to be_present # Sanity
```
awead: Want to put this in a `before` block? I'm quite liking the pattern now 😄

app/devise/ability.rb
```diff
@@ -26,6 +26,11 @@ def authenticated_permissions
   end
 
   def admin_permissions
-    can :manage, :all if current_user&.admin?
+    return unless current_user&.admin?
+
+    can :manage, :all
+    cannot :delete, Agent::Resource do |agent|
+      agent.member_ids.any?
+    end
```
awead: One of the changes that was introduced with Blacklight's access controls, was `ability_logic`, which is an array of abilities that are applied (in order) to a resource. This would enable you to make "cannot delete agents" its own kind of thing. Ex:

``` ruby
  self.ability_logic += %i[
    base_permissions
    authenticated_permissions
    admin_permissions
    cannot_delete_active_agents
  ]

  def cannot_delete_active_agents
    cannot :delete, Agent::Resource do |agent|
      agent.member_ids.any?
    end
  end
```
This way, if someone else has edit access later, but is not an admin, or some other unforeseen scenario, none of them would be able to delete that agent if it was being used in another resource.

awead: Order actually might not matter anyway, since a "can't" is a "can't" no matter where it occurs.

rschenk: Hmm yeah I see what you're saying. Supposing we have a "curator" role who is not an admin, but is allowed to CRUD one particular collection or whatever, we'd still like the "cannot" to be applied. I can do that

spec/cho/agent/agent_spec.rb
rschenk: I can, but then `agent` could no longer be a local variable, it'd have to be its own let. Which honestly is better.

Issue #975, opened 2019-05-09
Implementing private access on resources


awead:
When a user select "private access" on a resource, then only they will be able to read and edit it.

TODO:
* follow Hyrax paradigm and use a "private" to control this level of access
* capture the person who created the resource in the resource's metadata
* the attribute for the creator would not be stored in the data dictionary

awead:
System user doesn't seem to be correctly applied when a logged-in user is present.

PR #994, opened 2019-05-24
Enabling private access

awead:
## Description

Restricts collections to the users who have created them by adding a "system creator" attribute to the collection. The attribute is not present in the data dictionary but can be added to any collection or resource.

The system creator is always a member of read users, to ensure they have access to the resource. Additional changes were made to enforce read access rights defined in a collection's access level.

Note: Edit rights are not controlled. This will be dealt with in a later ticket.

Connected to #975 

## Changes

* refactored transactions into AccessControls module
* updated transactions to make changes to access control attributes
* access control attributes added change sets
* default system user or current user is assigned to all new resources


awead:
@rschenk Hold off on merging this. I need to discuss this more on Monday.

app/blacklight/access_controls/edit_ability.rb
```diff
@@ -0,0 +1,89 @@
...
+module AccessControls::EditAbility
+  extend ActiveSupport::Concern
+
+  def edit_permissions
+    can :edit, String do |id|
+      test_edit(id)
+    end
+
+    can :edit, SolrDocument do |obj|
+      cache.put(obj.id, obj)
```
rschenk: I have no idea what this `cache.put` thing is doing or why it's necessary, but is it a good idea for a policy object to have side effects that modify an external object/service?

app/cho/transaction/operations/access_controls/access_level.rb
```diff
@@ -0,0 +1,17 @@
...
+
+module Transaction::Operations::AccessControls
+  class AccessLevel
+    include Dry::Transaction::Operation
+
+    def call(change_set)
+      return Success(change_set) if change_set.try(:access_level).blank?
+
+      remaining_groups = change_set.read_groups - Repository::AccessLevel.names
+      change_set.read_groups = remaining_groups + [change_set.access_level]
```
rschenk: Necessary or advisable to use union operator `|` here instead of concat `+`?

app/cho/transaction/operations/access_controls/permissions.rb
```diff
@@ -0,0 +1,18 @@
...
+module Transaction::Operations::AccessControls
+  class Permissions
+    include Dry::Transaction::Operation
+
+    # @return [Valkyrie::ChangeSet]
+    # @note duplicate members can be removed by the underlying resource model
+    def call(change_set)
+      return Success(change_set) unless change_set.class.ancestors.include?(Repository::Access::ChangeSetBehaviors)
+
+      change_set.edit_users += [change_set.system_creator]
```
rschenk: Ditto, is it advisable to use a union `|` here instead of concatenation `+`. Maybe it doesn't really matter, but for future debugging it might be nice to not ever have duplicates in these types of lists?

Issue #726, opened 2018-11-13
Improve Solr Indexing of Dates


ntallman:
There are two solr fields for each date field in the metadata schema. A text-based Valkyrie field and a date-based Solr field. The Valkyrie field should be used for display purposes only and not indexed. The Solr field should be hidden from display and indexed.

Enhancement to #267

awead:
TODO:
* explore solutions for indexing EDTF dates in Solr so that they are actual dates and can be queried as such

Possible Solutions:
* two fields for each "date", start and end
* additional human-readable date

Ex:
2010-05 becomes "May 2010", start date 2010-05-01-00:00:00T (solr date), end date 2010-05-31-23:59:59T

awead:
1. Expand unambiguous dates into their ranges
    * determine which solr field to use, such as a date range
2. Ambiguous dates, such as "1984?" and expand it out into a range with a configured +/- tolerance value
    * tolerance values can be configured per unit, such as +/- 1 year, +/- 10 days

rschenk:
Hard problems such as how to handle intervals and sets, we are going to punt for now

awead:
Solr's date rang field https://lucene.apache.org/solr/guide/6_6/working-with-dates.html seems to support ranges expressed as edtf. For example:

> 2000-11 – The entire month of November, 2000.

PR #1002, opened 2019-05-30
Index EDTF dates as Solr DateRangeField

rschenk:
## Description

Parse EDTF dates and expand them to a Solr DateRangeField to allow them to appear in searches for a date that they contain.

**TODO:** 
* I've not figured out how to make Blacklight use this new field in the searches (I want to use `date_created_dtrsim` but it is using `date_created_tesim`)
* Still need to handle EDTF Sets?
* Uncertain dates are currently "exploded" to +/-1 least-significant-digit, which is how machinist tolerances are defined. Make this configurable at runtime?

Connected to #726

## Changes

* Reformat EDTF dates into the respective syntax for Solr DateRangeField
  * Certain dates of varying precision
  * Uncertain dates of varying precision
  * Unspecified dates of varying precision
  * Intervals with any combination of the above
* New indexer to add this field to a solr document if necessary
* New class to help abstract out the solr fields used by Blacklight's search bar


app/cho/indexing/edtf_date_range_formatter.rb
```diff
@@ -0,0 +1,110 @@
...
+                                         [:year]
+                                       end
+
+          potential_precision_levels.find { |level| date.specific?(level) }
+        end
+
+        # Idempotently parse strings into edtf dates
+        def parse_date_into_edtf(date)
+          if date.respond_to?(:edtf)
+            date
```
awead: We're only testing strings, so this line isn't getting tested. We could add one test case to fix that. But, to back it up a bit, I don't see any cases when there'd be something _other_ than a string. Does that seem right to you or is there a hidden case I'm not considering?

rschenk: I wasn't sure if we'd ever have a situation where there were not-strings given to this, but with these kinds of parsing methods, I like to make them idempotent when possible. For example `Date.edtf` itself is not idempotent, so if you give that method an edtf date, it has a conniption. That irritates me on a philosophical level. 😆 i'll add a test

Issue #976, opened 2019-05-09
Enable setting of access levels via CSV import


awead:
I should be able to specify the access level of a work during the CSV import.

TODO:
* make access level the data dictionary term `access_rights`
* provide a controlled vocabulary for the different levels: public, PSU access, and private
* provide a display transformation that takes the access level term in access rights and displays it as a more friendly term, ex. authenticated is displayed as "PSU Access" see #683 for details
* ensure the work edit forms use the data dictionary term

awead:
I'm punting on the display transformation because we currently don't have an easy way of implementing this. See #1013

awead:
Tying `:access_rights` to the application's implementation of access levels creates a unique link between the data dictionary and the application's own inner workings. This could have consequences if the term is changed or removed. We should make our future selves aware of that. See #1010

PR #1004, opened 2019-06-03
Setting access rights in CSV

awead:
## Description

Adds support for setting access rights in a CSV when adding new works to a collection.

Access rights is added to the data dictionary as a special kind of property that is linked to a `Repository::AccessLevel`. Setting the kind of access right will trigger behaviors such as limiting a resource (collection or work) to registered or private users.

Connected to #976 

## Changes

* adding access_rights to data dictionary
* Add radio_button field type used only for access rights at this time
* Adding validation and controlled vocabularies
* Replace access_level with access_rights where appropriate
* Making config changes to dev and prod data dictionaries

## Caveats

The access_rights property is tightly coupled with the repository access level concept, and functions will not work as expected unless this coupling is maintained. For example, changing access_rights to a different name would break these features. Future enhancements could make this coupling more transparent or flexible.

### DIFF
--- a/app/blacklight/catalog_controller.rb
+++ b/app/blacklight/catalog_controller.rb
@@ -180,14 +180,12 @@ class CatalogController < ApplicationController
     # solr fields to be displayed in the index (search results) view
     #   The ordering of the field names is the order of the display
     config.add_index_field 'work_type_ssim', label: I18n.t('cho.field_label.work_type')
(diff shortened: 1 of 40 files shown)


awead:
@rschenk not here, but I've ticketed it here #1010 which we can do in the nearer future.

Issue #1005, opened 2019-06-03
Only All Fields should search Extracted Text in Basic Search


ntallman:
Users expected a fielded search to only search that specific field. Only the All Fields basic search option should return results from extracted text.

Related: #804

PR #1015, opened 2019-06-04
Extracted text only added to "All Fields" search

rschenk:
## Description

Extracted text was previously being added to every Blacklight search, which was leading to some very unexpected results. Now, we only add extracted text to the query if "All Fields" is the selected search field.

Connected to #1005

### DIFF
--- a/app/blacklight/catalog_search_behavior.rb
+++ b/app/blacklight/catalog_search_behavior.rb
@@ -22,9 +22,16 @@ module CatalogSearchBehavior
 
     # the {!lucene} gives us the OR syntax
     def new_query
-      "{!lucene}#{internal_query(query_all_query_fields)} "\
-      "#{internal_query(search_metadata_from_file_sets)} "\
-      "#{internal_query(search_extracted_text_from_files)}"
+      query_parts = [
+        "{!lucene}#{internal_query(query_all_query_fields)}",
+        internal_query(search_metadata_from_file_sets)
+      ]
+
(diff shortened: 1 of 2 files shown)


spec/blacklight/catalog_spec.rb
```diff
@@ -113,11 +113,23 @@
...
+      end
+    end
+
+    context 'when searching a more specific field' do
+      it 'does not include the extracted text in the search' do
+        visit(root_path)
+        search_for 'words', in: 'Title'
+        within('#documents') do
+          expect(page).not_to have_link('Sample Extracted Text Work')
+        end
```
awead: @rschenk could we have an additional document added to the mix to improve the test contexts? I'm thinking add a work that has "words" in the title. The first context would expect both works to be in the search results, while the second context would only have the work with "words" in the title but not the other one.

app/blacklight/catalog_search_behavior.rb
```diff
@@ -22,9 +22,16 @@ def show_works_or_works_that_contain_files(solr_parameters)
 
     # the {!lucene} gives us the OR syntax
     def new_query
-      "{!lucene}#{internal_query(query_all_query_fields)} "\
-      "#{internal_query(search_metadata_from_file_sets)} "\
-      "#{internal_query(search_extracted_text_from_files)}"
```
mtribone: ah so there is the little bugger

rschenk:
(CI is failing due to ChromeDriver issues)

PR #1031, opened 2019-06-06
Updating to Valkyrie 2.0.0.RC10

awead:
## Description

Update Valkyrie to version 2.0.0.RC10. There will be an additional PR once the gem has reached official 2.0 status, but this captures the work done thus far on making CHO compatible with the updated gem. Future changes in later release candidates are likely to be minor. The biggest changes have to do with Valkyrie's dependent gems such as `dry-types` and `dry-struct` reaching a stable 1.0 status.

## Changes:

* Schema::Metadata

The behavior of core_fields and fields changed, and it seemed best to more clearly indicate that these are sets of ids. The attributes were changed to reflect their content, core_field_ids instead of core_fields and field_ids instead of fields.

* nil attributes

Calling .attributes on a model no longer returns nil values. If the value is nil, then the hash key is omitted. This only seemed to impact csv export of data dictionary fields, so we found a different way of returning all the attribute values for a given field.

* id sequences

Using sequence(:id) in FactoryBot was creating problems with Valkyrie's assignment of ids, so these were omitted and now Valkyrie just mints ids and we don't have to worry about sequencing them.

* update to dry-types 1.0

Updating to dry-types 1.0 affects our custom UseType class, which was reorganized to match the updated pattern.

* bug fixes

Some bugs surfaced with the update such as calling .members on a collection when it has no id, and an error in our File::Save operation that was setting the members relationship on the model instead of the change set.


awead:
Wait until after tomorrow's retro demo to merge this, just for sanity's sake.

PR #1038, opened 2019-06-12
Override Valkyrie::Persistence::Shared::JSONValueMapper::DateValue

awead:
## Description

Allows iso8601 dates that do not have time stamps, such as 1956-12-31, to be returned as strings and not DateTime objects. This allow for EDTF date-style processing at other points in the application.

Connected to #1035


awead:
@rschenk This is more of a "band-aid" approach that hopefully just unblocks us. It's definitely a bug in Valkyrie, which should not be casting unwitting strings into dates, but that's another matter. There's a more elegant solution to be had in there, at a later date.

Issue #951, opened 2019-04-10
The piwik id should change based on server


carolyncole:
It is hard coded to 2, which is the QA id number.  Stage is 5 and prod is 6.
https://github.com/psu-libraries/cho/blob/77f948eb8c7c38826a681f1ecc3275acc1e244f1/app/assets/javascripts/matomo.js#L12

awead:
TODO:
* make this a deployment-side configuration change, likely with Figaro environment variables
* do this how we're doing it Scholarsphere
* testing should include deploying to QA and staging to verify that it's working

rschenk:
I looked into doing this with Figaro, i.e. `application.yml` and indeed, it looks like this will be the place to do it. The `application.yml` file is one of the linked_files in the Capistrano, which I infer to mean that it's written by Chef and symlinked in after deploy. 

To do this, we'll need to add a key into the application.yml code and some minimal ERB to insert it into the javascript, but most of the effort will be involved talking to devops people to get the appropriate piwik ids into the respective Chef configs.

ntallman:
Confirmed to be correct in production!

PR #1042, opened 2019-06-12
piwik id is configurable by application.yml

rschenk:
## Description

Piwik id had been hard-coded to 2. It is now configurable via application.yml

In order for this to take effect, we will need to talk to the devops folks. The following key-value should be added to application.yml in chef on the following servers

```
# QA
piwik_id: '2'

# Stage
piwik_id: '5'

# Production
piwik_id: '6'
```

(Note that Figaro likes all values to be strings)

Connected to #951


awead:
I'll try pushing this branch to QA and see if it clears. If so, we can merge this.

awead:
Looks like this is working as expected. If I deploy to qa without the id in the yaml file, I get:

`Repository::Configuration::Error: piwik_id is a required key in application.yml`

If I add the value on the server, it deploys and is set correctly.

Issue #896, opened 2019-03-19
Elegant display of core fields in a collection


awead:
TBD.

mtribone:
Per #997 
- Only display public core fields with collection descriptive metadata
- Add "Collection Information" label to collection descriptive metadata

awead:
Blocked by #102

mtribone:
interface-inventory-cho.key

PR #1044, opened 2019-06-13
Collection Landing Page in two Columns without Resources

mtribone:
## Description

This is a small step in our Collection landing pages tickets. Still using the Blacklight partials and the search context (catalog) to display the show page. However, we have the title of the collection at the top and the description in the first column. Collection metadata goes into the second column.

Title and description are duplicated in the metadata output, but will be resolved in future work on the collection landing pages.

This paves the way for the representative image, browse button, searching the collection, etc.

Connected to #896 

## Changes

* Removes the tools partial from the display to give more room for the two column layout
* Adds the collection title to an `<h1>` at the top of the page
* Grabs the description field contents and displays it under the title in the first column
* Uses the default Blacklight show partials to outpout the metadata
* Adds a `<h2>` for the metadata

### DIFF
--- a/app/views/catalog/_show_collection_archival.html.erb
+++ b/app/views/catalog/_show_collection_archival.html.erb
@@ -1,3 +1,5 @@
+<%= document.description_data_dictionary_field.first %>
+
 <% @document.transformed_fields.each do |field| %>
   <% if field.present? %>
     <%= render partial: "shared/transformation_partials/#{field.display_transformation}",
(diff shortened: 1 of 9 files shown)


mtribone:
image

awead:
@mtribone seems like we need tests that verify the landing pages have what we currently expect them to. What about a new shared spec test for just archival and library collections that checks the contents of the landing page? The other feature tests can remain disabled.

PR #1051, opened 2019-06-20
Update schema.xml for Solr 8

awead:
## Description

Makes changes to the schema.xml to support Solr 8. Removes the StandardFilterFactory which has been deprecated and was a no-op, so removing it doesn't affect anything related to indexind.

Updates the different solr wrapper files so we can continue to use the solr_wrapper gem.


awead:
N.B. This configuration also works with Solr 7, so we could merge this and continue to use Solr 7 in our QA and prod environments. It will have no effect on indexing.

Issue #897, opened 2019-03-19
Representative imagery


awead:
Upload file to the public directory via ssh with the same name as the id, ex. pstma_mld.jpg. If no file exists, then nothing is shown. Put under one directory such as `public/collection_images`

Child of #483

awead:
TODO:
* make layout changes to display image
* upload files to `public/files` for display

awead:
Todo from June 21 retro:
* Path to image is wonky

awead:
Fixed in #1055

PR #1055, opened 2019-06-21
More robust collection image path generation

rschenk:
## Description

The original code I'd written to generate a URL for the collection representative images was a little too tricky for its own good and appeared to get confused with the symlinked directory in production. This new version is dumber, which in this case should be a good thing.

Connected to #897


awead:
Boom!
Screen Shot 2019-06-21 at 12 48 23 PM

PR #1056, opened 2019-06-24
Stimulus

awead:
## Description

Re-implements the UI for managing multiple fields using the simulus.js package and Jest for testing. Accomplishes the same feature set with less code and without any reliance on additional css classes. The only classes used are from Bootstrap.

## Changes

* creates `FieldsController` to do all the management
* uses Jest for unit tests
* linting and testing is done during add additional Travis build stage

## Preview

Screen Shot 2019-06-25 at 9 32 39 AM


spec/javascript/fields_controller_spec.js
```diff
@@ -0,0 +1,56 @@
+import { Application } from 'stimulus'
+import FieldsController from 'fields_controller'
+
+describe('FieldsController', () => {
+  beforeEach(() => {
+    document.body.innerHTML = `
+    <div class="form-group" data-controller="fields">
+      <label for="managed_field_id">
```
mtribone: Great. I couldn't readily tell from the code above if the label and input fields were connected. Looks like there will be an ID to match them up.

spec/javascript/fields_controller_spec.js
```diff
@@ -0,0 +1,56 @@
...
+      <div class="input-group">
+        <input type="text" id="managed_field_id">
+      </div>
+    </div>
+    <div class="form-group">
+      <label for="managed_field_id">
+        Unmanged Field
+      </label>
+      <div class="input-group">
+        <input type="text" id="unmanaged_field_id">
```
mtribone: If this is reflective of how the code is implemented, the input and label will not match up. Input has an ID of 'unmanaged_field_id' and the label has 'managed_field_id'.

app/javascript/controllers/fields_controller.js
```diff
@@ -0,0 +1,73 @@
...
+export default class extends Controller {
+  connect () {
+    let label = this.element.getElementsByTagName('label').item(0).innerText
+    for (let count = 1; count < this.inputGroups.length; count++) {
+      this.inputGroups.item(count).appendChild(this.removeButton)
+    }
+    this.element.appendChild(this.addButton(label))
+  }
+
+  // By default, Stimulus listens to click events on buttons and will execute this method.
```
mtribone: Click events _could_ be problematic for screen readers or keyboard users. We'll need to double check this in the UI and offer an alternative event listener if needed.

https://webaim.org/techniques/javascript/eventhandlers#onclick

app/javascript/controllers/fields_controller.js
```diff
@@ -0,0 +1,73 @@
...
+//     <div class="input-group">
+//       <input class="form-control" />
+//   </div>
+
+import { Controller } from 'stimulus'
+
+export default class extends Controller {
+  connect () {
+    let label = this.element.getElementsByTagName('label').item(0).innerText
+    for (let count = 1; count < this.inputGroups.length; count++) {
```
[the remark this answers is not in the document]
awead: @rschenk Sure! I didn't think of that, actually. However, I did want to skip the first element in the array so as not to render the "Remove" button on the first element. Looks like I could do this like:
``` javascript
this.inputGroups.forEach((input, index) => {
  if ( index === 0) {
    continue
  }
  else {
    input.item(index).appendChild(this.removeButton)
  }
})
```

spec/javascript/fields_controller_spec.js
awead: @mtribone Correct. Everything should the same as our current/standard HTML forms. There will, however, be multiple input fields per label:
``` html
<div class="form-group" data-controller="fields">
  <label for="work_submission_subject">
    Subject
  </label>
    
  <div class="input-group">
    <input aria-required="false" class="form-control" value="" type="text" name="work_submission[subject][]" id="work_submission_subject">
  </div>
  <div class="input-group">
    <input aria-required="false" class="form-control" value="" type="text" name="work_submission[subject][]" id="work_submission_subject">
    <button data-action="fields#remove" class="btn btn-outline-danger btn-sm">Remove</button>
  </div>
  <div class="input-group">
    <input aria-required="false" class="form-control" value="" type="text" name="work_submission[subject][]" id="work_submission_subject">
    <button data-action="fields#remove" class="btn btn-outline-danger btn-sm">Remove
    </button>
  </div>
  <button class="btn btn-outline-success btn-sm">Add another Subject</button>
</div>
```
I don't know what the accessibility ramifications are in this case.

spec/javascript/fields_controller_spec.js
awead: @mtribone Good catch. This is just bad test data. Really, this is just to illustrate that if the `date-controller="fields"` is missing from the div, then the controller won't do anything and will just leave the field alone. I should update the sample HTML though.

